PR #1230, opened 2013-06-18
Set default :jvm-opts in :base profile to :displace

ghost:
Leiningen's `:base` profile has the following defaults:

```
:jvm-opts ["-XX:+TieredCompilation" "-XX:TieredStopAtLevel=1"]
```

When the default behavior for `:jvm-opts` is additive, it can lead to
unexpected results if a project tries to add its own options which
conflict with the defaults. For example, the following setting:

```
:jvm-opts ["-XX:-TieredCompilation"]
```

Results in a JVM command line like this, in which the project's JVM
options are overridden by the defaults:

```
-XX:-TieredCompilation -XX:+TieredCompilation -XX:TieredStopAtLevel=1
```

This led to confusion on the part of people who were trying to run
benchmarks on some Clojure code. See the [mailing list discussion](https://groups.google.com/d/msg/clojure/LTtxhPxH_ws/jpQIg8PWMLsJ).

By setting `:displace` metadata on the default `:jvm-opts`, any
project configuration of `:jvm-opts` will replace the defaults.

Notes:

The wiki section on [Tiered Compilation](https://github.com/technomancy/leiningen/wiki/Faster#tiered-compilation) says one should use
`:jvm-opts ^:replace []` to turn off tiered compilation.

The [FAQ](https://github.com/technomancy/leiningen/blob/master/doc/FAQ.md) mentions that users can avoid tiered compilation by running
`lein with-profile production ...`.

### DIFF
--- a/leiningen-core/src/leiningen/core/project.clj
+++ b/leiningen-core/src/leiningen/core/project.clj
@@ -345,7 +345,8 @@
   profiles are active by default."
   (atom {:default [:base :system :user :provided :dev]
          :base {:resource-paths ["dev-resources"]
-                :jvm-opts tiered-jvm-opts
+                :jvm-opts (with-meta tiered-jvm-opts
+                            {:displace true})
                 :test-selectors {:default (with-meta '(constantly true)
                                             {:displace true})}
                 :target-path "target"


monokrome:
Apologies for the naivity but could someone give a small TLDR here of what the final decision was / how a Clojure noob can disable this? Maybe the only one (hope not) but I'm a bit confused

PR #1794, opened 2015-01-03
Translated Plugin page to Japanese.

k2n:
Hi, 

I translated the plugin page to Japanese. I asked proof reading at Qiita, a developer community site primarily for Japanese developers. 

http://qiita.com/k2nakamura/items/59890f1cedebe2812cca 

I'm going to translate other pages when I have a chance. Please accept the request. 

Thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 **/target
 *.bak
+*.swp
 *.class
 *asc
 *jar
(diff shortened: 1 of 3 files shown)


hyPiRion:
Sorry for not chiming in earlier – I assumed I had already replied to this one, but it appears not.

I'm positive to translated docs in general, but one of the challenges with docs in multiple languages is to find a way to tell users whether a translation is outdated or not (and, if possible, how outdated it is). That being said, it's not a discussion necessary to take in this PR. I'll file an issue on that later.

Thanks for the translation!

k2n:
Thanks for accepting the PR. Yes, it would be nice if we can come up with some mechanism to manage the version differences.

Issue #1743, opened 2014-10-25
uberjar uses uberjar profile even when a profile is explicitly specified


pangloss:
I build an uberjar in both a test configuration and a release configuration, so need to be able to run `lein with-profile test-uberjar uberjar`, but as of the most recent version of lein, that does not work and it always uses the uberjar profile.

hyPiRion:
Closed by #1801.

PR #1801, opened 2015-01-15
Fix for retaining :uberjar-name from profiles.

kephale:
Hello Friendly Leiningenites,

This is a fix for the bug where :uberjar-name isn't maintained from profiles. Tests pass and :uberjar-name is maintained when set inside a profile (albeit true that :uberjar-name is explicitly retained).

The gist of changes in this pull request: reusing some of the profiles code from jar.clj inside uberjar.clj, and re-associng uberjar-name after "preprocess-project" is called during jar-filename extraction. The idea is that uberjar-name wasn't being used from profiles inside uberjar.clj, and it was being removed during preprocess-project as well.

Cheers,
Kyle

### DIFF
--- a/src/leiningen/jar.clj
+++ b/src/leiningen/jar.clj
@@ -276,8 +276,7 @@ propagated to the compilation phase and not stripped out."
   (get-classified-jar-filename project (when uberjar? :standalone)))
 
 (defn get-jar-filename [project & [uberjar?]]
-  (get-jar-filename* (assoc (preprocess-project project)
-                       :uberjar-name (:uberjar-name project)) uberjar?))
+  (get-jar-filename* (preprocess-project project) uberjar?))
 
 (defn build-jar
   "Build a jar for the given project and jar-file."
--- a/src/leiningen/jar.clj
+++ b/src/leiningen/jar.clj
@@ -276,7 +276,8 @@ propagated to the compilation phase and not stripped out."
   (get-classified-jar-filename project (when uberjar? :standalone)))
 
 (defn get-jar-filename [project & [uberjar?]]
-  (get-jar-filename* (preprocess-project project) uberjar?))
+  (get-jar-filename* (assoc (preprocess-project project)
+                       :uberjar-name (:uberjar-name project)) uberjar?))
 
 (defn build-jar
   "Build a jar for the given project and jar-file."
--- a/src/leiningen/uberjar.clj
+++ b/src/leiningen/uberjar.clj
@@ -7,7 +7,8 @@
             [leiningen.core.project :as project]
             [leiningen.core.main :as main]
             [leiningen.core.utils :as utils]
-            [leiningen.jar :as jar])
+            [leiningen.jar :as jar]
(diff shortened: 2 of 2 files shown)


hyPiRion:
Hi there Kyle,

Thanks for the patch and the detailed description!

Is this a partial fix for #1743? Because I'm unable to reproduce this problem if there is no `:uberjar-name` entry in the `:uberjar`-profile. If I add
```clj
:buggy ^:leaky {:uberjar-name "new-uberjar-name.jar"}
```
to the profile map, things works as intended when running `lein with-profile [+]buggy uberjar`. Could you provide a minimal example that breaks? 

(Perhaps you're not aware of the new [`^:leaky` metadata](https://github.com/technomancy/leiningen/blob/master/doc/PROFILES.md#profile-metadata) that got added to 2.5.0? I certainly wouldn't immediately know about it if I weren't an active contributor here.)

kephale:
This does address #1743, and my use case is that I have a project with a bunch of profiles for different uberjars derived from the same project. 

Ah, yes, ^:leaky does the job. I wasn't aware of ^:leaky, and this certainly solves my own issue. Then again, when going through the uberjar.clj code, it wasn't particularly clear why the default behavior is different between uberjar and jar, which does not require ^:leaky.

kul:
@hyPiRion does this means that anything put in a profile without ^:leaky wont be respected in jar/uberjar/pom?

hyPiRion:
@kul, yes, you are correct. 

I'll have a better look at this patch later today – it seems to solve some of the merge order issues #1743 mentions.

kul:
@hyPiRion only problem with the patch is that it is limited to uberjar. But what if someone is using :aot true and :

hyPiRion:
@kul Right, the `:uberjar-name` seems actually to be superfluous, removing it had no change when using `^:leaky`.

Looking into the issue, it seems like this is the main culprit:

```clj
(let [project (project/merge-profiles project [:uberjar])
```

`:uberjar` is the last profile merged into the project, consequently all the `^:leaky` profile choices gets overridden by `:uberjar`. So I tried to reverse the order, such that `:uberjar` is the first profile merged into the project. In that case, all `^:leaky` profile choices should be preserved:

```clj
(let [{:keys [included-profiles excluded-profiles]} (meta project)
        ;; pick :uberjar values iff active leaky profiles doesn't define same values
        project (project/set-profiles project
                                      (cons :uberjar included-profiles)
                                      (remove #{:uberjar} excluded-profiles))
```

This does seem to resolve #1743, but I'd imagine @kephale's changes in `uberjar.clj` is a tad more robust.

I'm fine merging this one @kephale, if you unstage the `jar.clj` changes. With the `^:leaky` metadata attached, it looks like the `:uberjar-name` is correctly picked without it.

kephale:
@hyPiRion, okey doke. jar.clj is back to the way it was.

Thanks,
Kyle

hyPiRion:
Excellent, thanks! :)

PR #1804, opened 2015-01-20
Don't require port for drawbridge URLs

tpope:
REPL-y does not seem to require it, and I'm unaware of any other reason to mandate it.

### DIFF
--- a/src/leiningen/repl.clj
+++ b/src/leiningen/repl.clj
@@ -83,12 +83,14 @@
   (let [opt (str (first opts))]
     (if-let [sx (string-from-file opt)]
       (connect-string project [sx])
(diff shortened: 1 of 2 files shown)


tpope:
Travis failure appears to have been an issue retrieving dependencies and unrelated to the patch. Force pushing an amended commit to trigger a fresh run.

hyPiRion:
Yeah, travis tends to mess up downloading every now and then.

This seems alright to me, perhaps @trptcolin could verify that this is ok?

trptcolin:
I don't use use drawbridge (only reason I haven't commented). But seems fine to me.

hyPiRion:
Alright, then I don't see any reason to require ports.

Thanks!

PR #1807, opened 2015-01-23
Be explicit about array type.

brandonbloom:
I don't think this actually fixes [my problem][1], but into-array's single-argument semantics are evil.

[1]: https://gist.github.com/brandonbloom/d3598b6f2e4ef493ed56


brandonbloom:
I figured out my problem:

```
  :jvm-opts ^:replace [-"Xms512m" "-Xmx2g" "-server"]
```

Note the dash is before the quote! This means that a symbol array was getting created instead of a string array. In general, some better validation here would be welcome.

hyPiRion:
Nice catch, thanks.

PR #1808, opened 2015-01-24
Docs to fix gpg: can't query passphrase in batch mode in debian

mavbozo:
I use debian and when I did `lein deploy clojars` I got this error message:
```
gpg: can't query passphrase in batch mode
gpg: decryption failed: secret key not available
```

It turns out, I have to uncomment `use-agent` in `~/.gnupg/gpg.conf`.

I added the description of this problem and the solution in GPG.md and DEPLOY.md

### DIFF
--- a/doc/DEPLOY.md
+++ b/doc/DEPLOY.md
@@ -145,14 +145,15 @@ encrypted it. Due to a bug in `gpg` you currently need to use
 launches, but with `gpg-agent` you only have to enter your passphrase
 periodically; it will keep it cached for a given period.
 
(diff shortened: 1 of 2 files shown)


technomancy:
A couple things here: could you remove the duplication and just have it in one place and refer to it from the other? Also, I think the question of whether it uses the agent or not depends on which version of GPG is used; seems like that should be called out specifically.

Thanks.

mavbozo:
I remove the solution in DEPLOY.md because leiningen already give message to see `lein help gpg` when this error occurs. And, I also specifically mention gpg version 1.4.12

technomancy:
Thanks; looks good.

PR #1812, opened 2015-01-26
provide ability to supply a custom rendering function

yogthos:
Leiningen uses Stencil to render its templates, and that can be limiting in many situation. For example, I need to render Selmer templates that use `{{...}}` syntax and that conflicts with Stencil. I'm currently forced to use `with-redefs` to override the renderer as seen [here](https://github.com/yogthos/luminus-template/blob/b5753e158a0b8e1426ef7b87a0e3768bb698a264/src/leiningen/new/luminus.clj#L62). It would be nice to have the ability to supply a custom rendering function instead.

### DIFF
--- a/src/leiningen/new/templates.clj
+++ b/src/leiningen/new/templates.clj
@@ -124,14 +124,15 @@
   "Create a renderer function that looks for mustache templates in the
   right place given the name of your template. If no data is passed, the
   file is simply slurped and the content returned unchanged."
-  [name]
-  (fn [template & [data]]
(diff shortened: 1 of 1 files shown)


hyPiRion:
This looks good to me. Maybe add in a line or two in the docstring to explain what the `render-fn` argument is, and why you'd like to replace the default at times?

yogthos:
Good point, updated the docs. :)

hyPiRion:
Nice, thanks!

PR #1818, opened 2015-01-30
Add example for artifact classifiers. Closes #1589

### LINKED ISSUE #1589: Classifier for install
It would be nice if a `:classifier "xyz"` is available for classifying different profile installs to maven repos.

EDIT:
as @technomancy mentioned this feature is already present. The deployed artifact has `<classifier>xyz</classifier>` set in the .pom file but jar file is not suffixed with the classifier name. This can be reproduce by putting `:classifier "xyz"` in a profile `:xyz` and `:classifier "abc"` in another named `:abc`. I am using `lein with-profiles +abc install` to test this.

kul:


hyPiRion:
Could you add in the missing "is" in the sentence `If a keyword it looked up from :profiles`? Otherwise, this looks fine.

kul:
Squashed in a new commit.

hyPiRion:
Thanks :)

PR #1822, opened 2015-02-05
Fixes technomancy/leiningen/#1308

mydoghasworms:
Change the powershell option to use the default system proxy in Windows using solution from here (so if you are already authenticated with the proxy, it authenticates you in PowerShell): http://stackoverflow.com/a/25082595/274354

### DIFF
--- a/bin/lein.bat
+++ b/bin/lein.bat
@@ -156,7 +156,7 @@ if NOT ERRORLEVEL 1 (
 )
 call powershell -? >nul 2>&1
 if NOT ERRORLEVEL 1 (
-    powershell -Command "& {param($a,$f) (new-object System.Net.WebClient).DownloadFile($a, $f)}" ""%2"" ""%1""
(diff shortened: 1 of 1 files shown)


hyPiRion:
From the little I know, this seems alright. Could anyone on windows verify that this looks okay (and, if possible, that it works)?

winks:
I can't tell you that it works with a proxy because I don't have one, but what I tested now on Win7:

  - old code, "official" cert (in browser by vendor): works
  - new code, "official" cert (in browser by vendor): works

  - old code, "cacert" cert (manually installed): works
  - new code, "cacert" cert (manually installed): works

  - old code, "self-signed" cert (browser shows error): doesn't work, powershell error
  - new code, "self-signed" cert (browser shows error): doesn't work, powershell error

so apart from the proxy stuff it doesn't seem to make anything worse than it is.

hyPiRion:
@winks – thanks for the verification. I'll assume this is alright then.

@mydoghasworms, thanks for the fix :)

Issue #1800, opened 2015-01-15
Certificate Error On Windows


mike-thompson-day8:
The fix for #1733 broke `lien.bat` on Windows  (#126 seems related)

Detailed reports: https://groups.google.com/forum/#!topic/clojure/NTuFH0YGDeo 

We've investigated a bit and found that **neither** `wget` or `curl` work on Windows without the insecure command-line params. 

As an experiment, we removed access to BOTH `wget` and `curl` from PATH and found that `powershell` works beautifully. 

So the choice is:
   1.  Put the insecure command-line parameters back in  (doesn't seem like an option)
   2.  Remove use of `wget` and `curl`, and insist that `powershell` is installed.

bruceadams:
Requiring Powershell seems reasonable. Hasn't Powershell been included with every Windows release for the last eight or ten years now?

mike-thompson-day8:
Powershell is been built into Windows from Win7 onwards (released July 22, 2009).  It also seems to have been backported to WinXP via service Pack 3.

Source: "Versions" in http://en.wikipedia.org/wiki/Windows_PowerShell

So, to be without powershell,  a developer would (1) need to be using XP -- ie be using a PC that was over 6 years old plus (2) they must have chosen to not install service pack 3. 

Using `powershell` the first choice downloader seems completely reasonable and seems the right response. That has to handle 99.9% of cases.  

The only question then is:  do you want to continue to use the insecure version of `wget` and `curl` as a fall back for anyone on XP?

technomancy:
> The only question then is:  do you want to continue to use the insecure version of `wget` and `curl` as a fall back for anyone on XP?  

No; if we include any fallback it should be to secure invocations. If anyone has trouble they can fix their CA store or download in a browser manually.

kumarshantanu:
Powershell in `lein.bat` has worked well (commenting out _wget_ and _curl_ invocations) for me. It can be installed on Windows XP/Vista and comes pre-installed on Windows 7 and higher.

hyPiRion:
If it's easier to install PowerShell (which I suspect?) than wget/curl and that it also handles CA stores better, then I'd recommend to download via PowerShell by default and fallback on curl, then wget. It seems like wget is the one with the most issues, and I've heard of people that fixed this by just using curl instead.

djpowell:
[...]

PR #1831, opened 2015-02-11
Reorder the possible HTTP proxies to use to download Leiningen from GitH...

fgiasson:
...ub. The new sequence is: PowerShell, Curl and then Wget.

Fixes issue #1800 as experienced by multiple users.

### DIFF
--- a/bin/lein.bat
+++ b/bin/lein.bat
@@ -141,6 +141,11 @@ if NOT "x%HTTP_CLIENT%" == "x" (
     %HTTP_CLIENT% %1 %2
     goto EOF
 )
+call powershell -? >nul 2>&1
+if NOT ERRORLEVEL 1 (
+    powershell -Command "& {param($a,$f) $client = New-Object System.Net.WebClient;  $client.Proxy.Credentials =[System.Net.CredentialCache]::DefaultNetworkCredentials; $client.DownloadFile($a, $f)}" ""%2"" ""%1""
+    goto EOF
+)
 call wget --help >nul 2>&1
 if NOT ERRORLEVEL 1 (
     call wget -O %1 %2
@@ -152,11 +157,6 @@ if NOT ERRORLEVEL 1 (
     set CURL_PROXY= 
     if NOT "x%HTTPS_PROXY%" == "x" set CURL_PROXY="-x %HTTPS_PROXY%"
     call curl %CURL_PROXY% -f -L -o  %1 %2
-    goto EOF
-)
-call powershell -? >nul 2>&1
-if NOT ERRORLEVEL 1 (
-    powershell -Command "& {param($a,$f) $client = New-Object System.Net.WebClient;  $client.Proxy.Credentials =[System.Net.CredentialCache]::DefaultNetworkCredentials; $client.DownloadFile($a, $f)}" ""%2"" ""%1""
     goto EOF
 )
 goto NO_HTTP_CLIENT
--- a/bin/lein.bat
+++ b/bin/lein.bat
@@ -136,6 +136,7 @@ if "x%JVM_OPTS%" == "x" set JVM_OPTS=%JAVA_OPTS%
 goto RUN
 
 :DownloadFile
+set LAST_HTTP_CLIENT=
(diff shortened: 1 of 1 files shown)


winks:
+echo It is possible that the download failed due to "curl", 
+echo "wget" or "powershell" inability to retreive GitHub's certificate.

should probably be something like

+echo It is possible that the download failed due to "powershell", 
+echo "curl" or "wget" not being able to retrieve GitHub's certificate.

And removing powershell from path doesn't sound like an easy task, and the order is wrong.

I think the cleanest solution would be to add an optional environment variable to override powershell, if it *really* doesn't work.

Sorry if this sounds overly pedantic :)

hyPiRion:
We have the `HTTP_CLIENT` environment variable that you're able to set. It's possible to specify a way to download insecurely (with emphasis on it being insecure) by e.g. setting it to  `call curl --insecure -f -L -o`.

Perhaps making a powershell example of setting that variable would be an idea?

winks:
oh right, forgot about this and that it's supposed to work on windows too.

fgiasson:
@hyPiRion Yeah that is right. However, any idea how to setup `HTTP_CLIENT` to work with the powershell? Particularly since the parameters appears to be in reverse order?

Also, I can't find the code where it says to try to use the `HTTP_CLIENT` environment variable to call `curl` or `wget`. Any pointers?

hyPiRion:
@fgiasson I don't have enough experience with bat/PowerShell to know how that's done unfortunately.

If you wonder about the insecure example, that's just the old insecure code which was removed in d63b2fa40f7f2042c80b1da676667145ca6a1a3c. There isn't any example of how to use `HTTP_CLIENT` as an env variable, which is why I suggested to make one for it.

For reference, the source for `HTTP_CLIENT` usage is here: https://github.com/technomancy/leiningen/blob/772d38342f7beaa06dadb303f76cad77111659d4/bin/lein.bat#L140-L143

fgiasson:
@hyPiRion Thanks for the reference. Yeah, I don't neither for PowerShell. Anybody else as a clue?

winks:
I think it's fine as it is

I just played around with your change merged on windows and it seems to work as expected:

  1. powershell available
    1. works -> fine
    1. doesn't work -> error msg
      1. user does `set HTTP_CLIENT=wget or curl`
        1. -> works -> fine
        1. -> doesn't work -> worst case. maybe the user will figure it out, maybe they will seek help  1. powershell not available
    1. lein.bat tries wget, then curl
      1. works? -> ok
      1. doesn't work? -> see worst case again


So I think the worst case can't really be solved in a nice automated way that's *better* than now.

TLDR: shouldn't be worse, can only better, let's go. (unless I missed something)

fgiasson:
@winks, @hyPiRion,

Check this new iteration: https://github.com/fgiasson/leiningen/commit/9b60ea0332002dcf43d7dc55f96c4c2112c0cc3c

Should be the behavior outlined by @winks. I agree with that proccess.

However, please test that. I am currently getting that PowerShell error:

```
Exception calling " downloadFile " with "2" argument ( s) : "Demand
was aborted : The connection was unexpectedly terminated ".
The character line: 1: 145
+ ... rkCredentials; $client.DownloadFile($a, $f)}
"https://github.com/technomancy/lei ...
+                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], MethodInvocationException
    + FullyQualifiedErrorId : WebException
```

And I am not sure why.

winks:
@fgiasson Did you change `2.5.2-SNAPSHOT` to `2.5.1` in line 5? because otherwise it can't work.

I just downloaded your HEAD from your branch and just with that change 2.5.1 downloaded perfectly on Win 7 Home Premium SP1.

fgiasson:
@winks nah, that was the actual issue thanks :)

Working perfectly here.

So, is that patch now acceptable to everyone?

hyPiRion:
Yeah, this is super :) Thanks!

fgiasson:
perfect thanks!

PR #1832, opened 2015-02-12
Fix .hgignore in new with regex. Fixes #1815

### LINKED ISSUE #1815: new and dot hgignore
It was nice to see support for dot hgignore added, but some of the rules are wrong. For example "/.lein-*" should be ".lein-*". Also, some should be rooted to avoid pruning subdirectories and files. For example, "target/**" should be in a "regexp" syntax section as "^target/".

> Neither  glob  nor regexp patterns are rooted. A glob-syntax pattern of
       the form *.c will match a file ending in .c in  any  directory,  and  a
       regexp pattern of the form \.c$ will do the same. To root a regexp pat‐
       tern, start it with ^.

Also, it would be nice to see some emacs support, for example

    syntax: glob
    *~				# emacs backup files

    syntax: regexp
    \.\#.*			# emacs cache files
    \#.*\#			# emacs autosave files

winks:
Just fixing the issues, no emacs changes.

### DIFF
--- a/resources/leiningen/new/app/hgignore
+++ b/resources/leiningen/new/app/hgignore
@@ -1,12 +1,14 @@
 syntax: glob
-target/**
-classes/**
-checkouts/**
 pom.xml
 pom.xml.asc
 *.jar
 *.class
-/.lein-*
-/.nrepl-port
 .gitignore
 .git/**
+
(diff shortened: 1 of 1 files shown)


hyPiRion:
Thanks.

Issue #1835, opened 2015-02-17
Can't use :project/source-paths in aliases


arrdem:
For [lein-grim](https://github.com/clojure-grimoire/lein-grim) I'm attempting to make use of the following alias:

```clojure
:aliases {"grim" ["run" "-m" "grimoire.doc"
                  ,,:project/groupid
                  ,,:project/artifactid
                  ,,:project/version
                  ,,:project/source-paths]}
```

It's not clear to me how, but I'm getting the following exception:

```shell
% lein grim src doc/
Exception in thread "main" java.lang.ClassCastException: java.lang.String cannot be cast to clojure.lang.IFn, compiling:(/tmp/form-init1746183400574885756.clj:1:72)
        at clojure.lang.Compiler.load(Compiler.java:7207)
        at clojure.lang.Compiler.loadFile(Compiler.java:7151)
        at clojure.main$load_script.invoke(main.clj:274)
        at clojure.main$init_opt.invoke(main.clj:279)
        at clojure.main$initialize.invoke(main.clj:307)
        at clojure.main$null_opt.invoke(main.clj:342)
        at clojure.main$main.doInvoke(main.clj:420)
        at clojure.lang.RestFn.invoke(RestFn.java:421)
        at clojure.lang.Var.invoke(Var.java:383)
        at clojure.lang.AFn.applyToHelper(AFn.java:156)
        at clojure.lang.Var.applyTo(Var.java:700)
        at clojure.main.main(main.java:37)
Caused by: java.lang.ClassCastException: java.lang.String cannot be cast to clojure.lang.IFn
        at user$eval5.invoke(form-init1746183400574885756.clj:1)
        at clojure.lang.Compiler.eval(Compiler.java:6768)
        at clojure.lang.Compiler.eval(Compiler.java:6758)
        at clojure.lang.Compiler.load(Compiler.java:7195)
        ... 11 more
```

It seems to me that the `:project/source-paths` value stringifies into a `()` list which when string formatted and passed to the child process becomes interpreted as the invocation of the string in the 1st position giving the resulting error. As this is blocking on my lein-grim work I'm gonna put some time into tracking it down but as of my last look at the lein codebase it wasn't clear how to attack solving this. Feedback would be appreciated.

hyPiRion:
[...]

PR #1837, opened 2015-02-18
Normalize most sequence arguments to vectors

arrdem:
This patch attempts to address technomancy/leiningen#1835 by examining
arguments and providing list to vector rewriting for a class of inputs
(those lists which are not traditional symbol named function
invocations). This would allow the use of sequence result alias
arguments like `:project/source-paths` which otherwise serialize to
list expressions and fail on the evaluation side.

### DIFF
--- a/src/leiningen/run.clj
+++ b/src/leiningen/run.clj
@@ -14,71 +14,84 @@
     (symbol given)
     (symbol (name given) "-main")))
 
+(defn- invocation? [arg]
+  (and (list? arg)
+       (or (symbol? (first arg))
+           (invocation? (first arg)))))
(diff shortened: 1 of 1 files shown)


hyPiRion:
I did some changes to the normalization phase: I noticed that the run form injects `args` at multiple positions, which caused havoc when forms caused side effects. So I decided that it made more sense to just quote the args.

You can still get the same functionality by calling `eval` on the arguments passed to the provided function, though.

arrdem:
Great. I don't personally have a use for that, but I figured that'd be the solution that made the most sense. Thanks!

PR #1838, opened 2015-02-19
Added Artifactory to the list of repository managers

jbaruch:


hyPiRion:
Thanks!

(Sorry for the delay here, I just wanted to verify that the ones mentioned is the repository manager "trio")

jbaruch:
Thanks! [Here is more info and comparison of the "trio"](http://docs.codehaus.org/display/MAVENUSER/Maven+Repository+Manager+Feature+Matrix).

PR #1842, opened 2015-02-24
Move read-checkouts to project module and make it public

avasenin:
It's needed for plugins which behaviour depends on checkouts. e.g. cljs-build auto should track changes in checkouts too.

I tried to solve circular dependency between classpath and project modules but without success. If you give me some advice then I could do this in scope of this task to remove this circular dependency.


hyPiRion:
Sorry for the late response here.

I'll have a look at this one before 2.5.2 is out – I'm initially positive to this change, just need to verify that it doesn't cause any problems.

hyPiRion:
Looks good, thanks!

The circular dependencies aren't possible to clear up without breaking the API, so they are going to stay until 3.0 is out.

PR #1845, opened 2015-02-25
bail-out logic

noisesmith:
makes lein bail out immediately rather than compiling the whole project before complaining


technomancy:
Looks good; thanks.

PR #1847, opened 2015-03-04
Update TUTORIAL.md

kevincolyar:
Updates checkouts instructions regarding absolute vs. relative symlinks.

### DIFF
--- a/doc/TUTORIAL.md
+++ b/doc/TUTORIAL.md
@@ -280,7 +280,7 @@ create a directory called `checkouts` in the project root, like so:
         `-- my_stuff
             `-- core_test.clj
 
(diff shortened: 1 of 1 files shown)


kevincolyar:
Scratch that.  I thought this had solved my issue with my checkout's project.clj not being found but no luck.

hyPiRion:
Is the problem you're having worthy of an issue?

kevincolyar:
No, I figured out my problem was trying to include "checkouts" in my :source-paths so that "lein cljsbuild auto" would pick up the changes.

Found out the fix for this is in an open pull request: https://github.com/emezeske/lein-cljsbuild/pull/374

PR #1851, opened 2015-03-06
cleaxure has gone too far

tcrayford:
cleaxurely we should prevent these kinda puns without some kinda eaxverride environment variable


technomancy:
When will the madness end?

kevinburke:
when will github add a fave button

hyPiRion:
Dangit, I forgot to merge this on April 1st :(

tcrayford:
wanna just merge it now then?

hyPiRion:
Yeah, why not

PR #1861, opened 2015-03-18
Bump nREPL dep to 0.2.8

bbatsov:
Title says it all.


hyPiRion:
Thanks for this one too :)

bbatsov:
Btw, it'd be really great if a version `lein` using 0.2.8 is released soon. 2.5.1 bundled 0.2.6 and a lot of bugs were fixed since then. Most people start their repls using `lein repl` and don't really know how to update `nREPL` by itself.

Guess this is up to @technomancy, but I still wanted to mention it.

PR #1870, opened 2015-04-01
Allow global leinrc file

pksunkara:
I want to global leinrc instead of creating the same for each user.

### DIFF
--- a/bin/lein
+++ b/bin/lein
@@ -98,7 +98,7 @@ done
 
 export LEIN_HOME="${LEIN_HOME:-"$HOME/.lein"}"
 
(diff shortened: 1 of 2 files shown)


hyPiRion:
Hi there,

Could you reorder the filenames so that they are in the order `"/etc/leinrc" "$LEIN_HOME/leinrc" ".leinrc"`?

The order shouldn't matter, except where people intentionally want to override things defined "lower down" in the system.

Otherwise this looks good.

pksunkara:
@hyPiRion Updated.

hyPiRion:
Excellent, thanks!

PR #1881, opened 2015-04-14
Provides another option for `lein new` to specify template version

athos:
This PR proposes another option `--template-version` for `lein new` to specify the template version to be used.

The version of Leiningen templates is often tied up with its dependent libraries. In some cases, however, we would want to use a specific version of them. It's convenient Leiningen provides a way to specify the version of the template to be applied.


hyPiRion:
Looks good, thanks!

PR #1886, opened 2015-04-23
Refactor main/drop-partial-args

oneeman:
The refactoring was prompted by the `TODO`.  I believe it's a little better now.  This function is private and so is not tested directly.  I did add some tests which exercise partial application of task arguments with `resolve-task`, as well add a dummy project which has both fixed and `&` arguments, since there didn't seem to be one.

This is my first dip into the Leiningen codebase and I'm pretty new to Clojure as well, so please let me know if any of the terminology is off or whatever else.


hyPiRion:
Hi there,

I'm terribly sorry for the late response here. This seems fine, thank you!

If one really wanted to test private functions, one could get it through its var (`#'namespace/internal-function`), but I don't think it's necessary here.

oneeman:
Good to know, thanks!

PR #1906, opened 2015-05-19
Address two questions about checkout dependencies

marick:
Two questions I've had, couldn't answer from the leiningen documentation, but found an answer to somewhere else.

### DIFF
--- a/doc/TUTORIAL.md
+++ b/doc/TUTORIAL.md
@@ -317,7 +317,7 @@ hierarchy doesn't need to appear in the way `commons` is actually laid out on di
 After you've updated `:dependencies`, `lein` will still need to be able
 to find the library in some repository like clojars or your `~/.m2`
 directory.  If `lein` complains with a message like "Could not find artifact suchwow:jar:0.3.9",
-it's possible that `project.clj` and `suchwow/project.clj` have different version numbers.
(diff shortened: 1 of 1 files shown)


technomancy:
Looks good.  I wonder if the second question could be clearer if we just specified that the directory name in the `checkouts/` directory is completely arbitrary. (Been a while since I've looked at this code, but I think this is true.) Then it's quite clear it doesn't matter what the group-id is.

marick:
I should probably then edit the preceding text. OK?

technomancy:
Sure, that'd work. Thanks.

marick:
I've finished the edits. Note: I left the following alone, because I don't understand it:

"If you change the
dependencies of a checkout project you will still have to run `lein
install` and restart your repl; it's just that source changes will be
picked up immediately."

hyPiRion:
This looks good to me, so I merged it in. I did a small fix on the dependency vector, but otherwise it looked fine.

The sentence you quote attempts to tell you that if you change the dependencies in e.g. the `suchwow` project, then that doesn't propagate down to the project that uses `suchwow` through checkout dependencies. In that case, you have to do `lein install` (and restart the repl) to get those new dependencies downstream.

PR #1912, opened 2015-06-03
Include more details for testing plugins locally in PLUGINS.md

timlinquist:
I spent about an hour trying to figure this out before a colleague passed this advice onto me. Seems important to codify it in the plugins.md

### DIFF
--- a/doc/PLUGINS.md
+++ b/doc/PLUGINS.md
@@ -31,7 +31,7 @@ See the `lein-pprint` directory
 for a sample of a very simple plugin.
 
 ### Local development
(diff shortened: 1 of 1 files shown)


technomancy:
This looks good, except the first sentence isn't really true of very many plugins. Usually you can't test the plugin on itself, and many plugins don't contain any tasks at all. If you can re-word it to be more general that would help.

timlinquist:
@technomancy Updated. Thanks for the project. Dramatically simplified the workflow for me learning clojure akin to the usefulness rake in the Ruby world.

technomancy:
Glad it's helpful for you. Merged this change in; thanks.

Issue #750, opened 2012-08-27
Cannot use .lein-classpath for plugin development


Adirelle:
Despite what is indicated in https://github.com/technomancy/leiningen/blob/master/doc/PLUGINS.md#writing-a-plugin, it is not possible to add the plugin sources in the file .lein-classpath of a test project to avoid resintalling it on every change. Once the plugin is added to the test project, Leiningen 2.x tries to use the artifact from local repository or to download it.

Test case:
1. create a new plugin project lein-testme, using `lein new plugin lein-testme`,
2. create a test project, using `lein new test-lein-testme`,
3. add lein-testme as a required plugins of test-lein-testme `:plugins [[lein-testme "0.1.0-SNAPSHOT"]]`,
4. add lein-testme source path to the classpath of lein for test-lein-testme  : `echo ../lein-testme/src >test-lein-testme/.lein-classpath`,
5. try to launch lein in test-lein-testme :

```
Could not find metadata lein-testme:lein-testme:0.1.0-SNAPSHOT/maven-metadata.xml in central (http://repo1.maven.org/maven2)
Could not find metadata lein-testme:lein-testme:0.1.0-SNAPSHOT/maven-metadata.xml in clojars (https://clojars.org/repo/)
Could not find artifact lein-testme:lein-testme:pom:0.1.0-SNAPSHOT in central (http://repo1.maven.org/maven2)
Could not find artifact lein-testme:lein-testme:pom:0.1.0-SNAPSHOT in clojars (https://clojars.org/repo/)
Could not find artifact lein-testme:lein-testme:jar:0.1.0-SNAPSHOT in central (http://repo1.maven.org/maven2)
Could not find artifact lein-testme:lein-testme:jar:0.1.0-SNAPSHOT in clojars (https://clojars.org/repo/)
Check :dependencies and :repositories for typos.
It's possible the specified jar is not in any repository.
If so, see "Free-floating Jars" under http://j.mp/repeatability
Could not resolve dependencies
```

PR #1917, opened 2015-06-13
PLUGINS.md: Clarify some things

rmoehn:
While trying to develop a Leiningen plugin I stumbled over two small
things, which I think make other people stumble as well:

 - Adding the path to my plugin's src directory to .lein-classpath
   didn't help anything. Leiningen still loaded the coded that I had
   installed to the Maven repository and ignored the local changes.
   Putting :provided or :resolve or :classpath after the plugin
   coordinates as suggested in issue #750 didn't help either. After
   banging my head against this, I got the idea to remove the plugin
   entry completely and it worked. Not sure if this is the intended
   behaviour, though. Add a step-by-step guide to getting what you want
   anyway.

 - It wasn't at all clear to me how to do the subtasks thing. The
   sentence about the :subtask metadata is a bit garbled and I thought
   there might be a mechanisms that invokes subtasks for me. I couldn't
   find anything about that, though. A quick search took me to
   https://github.com/devth/lein-worker/blob/master/src/leiningen/worker.clj,
   which showed the relevant bits. Add an example, because that's what
   people understand.

Discussion:

 - The new place of the paragraph for emitting output is not very good,
   but it's better than leaving it at the bottom of the newly expanded
   local development section, where nobody would see it.

 - Renaming "Documentation" to "Documentation and subtasks" is a rather
   thin patch. – Instructions on how to invoke subtasks don't really
   belong in a section about documentation. – However, I'm too lazy to
   do a lot of restructuring and I think people will still find what
   they're looking for.

### DIFF
--- a/doc/PLUGINS.md
+++ b/doc/PLUGINS.md
@@ -30,6 +30,10 @@ See the `lein-pprint` directory
 [in the Leiningen source](https://github.com/technomancy/leiningen/tree/stable/lein-pprint)
 for a sample of a very simple plugin.
 
+When emitting output, please use `leiningen.core.main/info`,
+`leiningen.core.main/warn`, and `leiningen.core.main/debug` rather than
+`println` since these will respect the user's output settings.
+
 ### Local development
 When you're ready to test your plugin in a separate project you can include it via the following (example a plugin named sample-plugin):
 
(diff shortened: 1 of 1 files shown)


hyPiRion:
Sorry for the delay here.

I looked over the changes, and I found them good. Only thing I'd like to discuss is the sentence "This might seem strange, but it won't work otherwise." in the fifth step. The reason Lein will pick up the `lein install`ed version instead of the classpath one when you have the plugin installed, is because Lein loads the plugins after `.lein-classpath` had been loaded, which overrides other versions on the classpath. Perhaps it would be better to write that instead (in a more concise way)?

rmoehn:
Yeah, sure. Changed it. I'm not quite content with the formulation, but it should do.

hyPiRion:
Yeah, that's a good enough explanation for me.

Thanks!

Issue #687, opened 2012-07-17
`shasum` not available on CentOS 5, Fedora 16


haus:
Issue #458 and commit e0c6e486 changed `sha1sum` to `shasum`, which isn't available on CentOS or Fedora 16.

icambron:
@technomancy I know this ticket is ancient, but the use of `shasum` is back: https://github.com/technomancy/leiningen/blob/ee57b19a5daae0687f22c7aba0da55538366664f/bin/lein#L308, and I had the same issue on CentOS

Leonidas-from-XIV:
I had the same issue on Fedora 22 and Leiningen 2.5.0. I believe this issue should be reopened.

hyPiRion:
Closed by #1918.

PR #1918, opened 2015-06-14
Platform independent SHA

Leonidas-from-XIV:
This addresses the problem in #687. Not all systems have `shasum`, some have `sha1sum` instead. Feel free to suggest other solutions, this was the first that came to my mind.

(`$SUM` was already used before in the history of leiningen, so I just revived that variable name)

### DIFF
--- a/bin/lein
+++ b/bin/lein
@@ -23,11 +23,10 @@ else
   cygwin=false
 fi
 
-if command -v shasum >/dev/null 2>&1; then
-    SUM="shasum"
-elif command -v sha1sum >/dev/null 2>&1; then
-    SUM="sha1sum"
-fi
+function command_not_found {
+    >&2 echo "Leiningen coundn't find $1 in your \$PATH ($PATH), which is required."
+    exit 1
+}
 
 function make_native_path {
     # ensure we have native paths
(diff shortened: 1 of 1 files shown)


hyPiRion:
This looks good, only thing I'd like to have is a proper error message if neither `shasum` or `sha1sum` is found. Currently it just errors because `$SUM` is empty. I guess this could be solved with something like this:

```bash
function command_not_found {
    echo "Lein couldn't find $1 on the path, which is needed to be ran"
    exit 1
}

## Then the sum function
if command -v shasum >/dev/null 2>&1; then
    SUM="shasum"
elif command -v sha1sum >/dev/null 2>&1; then
    SUM="sha1sum"
else
    SUM=command_not_found "shasum or sha1sum"
fi
```

Does that change look reasonable to you?

winks:
@Leonidas-from-XIV and I talked about this and in the end we both liked this version better, because this would let lein die *every time* both can't be found - but you only need this for fast trampoline.

Maybe the best solution is to put that `exit` inside the if for **fast trampoline**

Leonidas-from-XIV:
Yes, I agree, we should just put an error message in the fast trampoline case. I'll revise the PR.

hyPiRion:
@winks: That's the reason I wrapped the error message along with `exit 1` inside a function, instead of calling it immediately.

winks:
right.. you confused me there :)

Leonidas-from-XIV:
So, this combines both.

(Actually, trampolines do not really seem to work for me, but this looks like a different issue and shall be adressed in a future PR)

hyPiRion:
Yeah, this looks fine. Thanks! :)

PR #1919, opened 2015-06-19
Escape variable so script does not break on space in username

marthall:
Running `lein` on OSX Yosemite 10.10.3 breaks with error `/usr/local/bin/lein: line 9: [: too many arguments`. Separating the `whoami` command in a variable and escaping this solves the problem.

### DIFF
--- a/bin/lein-pkg
+++ b/bin/lein-pkg
@@ -6,9 +6,7 @@
 
 export LEIN_VERSION="2.5.2-SNAPSHOT"
 
-export USR=`whoami`
-
-if [ "$USR" = "root" ] && [ "$LEIN_ROOT" = "" ]; then
+if [ "`whoami`" = "root" ] && [ "$LEIN_ROOT" = "" ]; then
     echo "WARNING: You're currently running as root; probably by accident."
     echo "Press control-C to abort or Enter to continue as root."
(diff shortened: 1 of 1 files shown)


bin/lein-pkg
```diff
@@ -6,7 +6,9 @@
 
 export LEIN_VERSION="2.5.2-SNAPSHOT"
 
-if [ `whoami` = "root" ] && [ "$LEIN_ROOT" = "" ]; then
```
bruceadams: The `whoami` can be quoted, like this:

    if [ "`whoami`" = "root" ] && [ "$LEIN_ROOT" = "" ]; then

I find this code easier to read. Introducing the USR environment variable leaves me wondering if it is used elsewhere.

hyPiRion:
dang, nice catch.

As @bruceadams mentioned, I think this would be clearer with either ``"`whoami`"`` or `"$(whoami)"` instead of adding a new variable. Will be happy to merge it in after that (or if there's a reason `USR` is used).

marthall:
You are completely right! My bash skills are not as good as I want :smile: . It should be fixed now.

hyPiRion:
Looks good, thanks! :)

Issue #1827, opened 2015-02-10
Reader Conditionals


hyPiRion:
<http://dev.clojure.org/display/design/Reader+Conditionals> contains information about upcoming changes in Clojure (presumably 1.7). Right now, the only thing that's of importance seems to be supporting reading/compiling `.cljc` files.

Enabling/disabling feature sets would only require the user to update the `:jvm-opts`, so there should be no issues there. However, merging these options would be messy (#878 might remedy this problem).

jenanwise:
Right now lein completely ignores .cljc files (e.g., `lein test` will not pick them up) because bultitude does not recognize them. I added partial support in of Raynes/bultitude#27 but leiningen's use of `bultitude.core/path-for` will have to modified/removed since the 1-to-1 namespace-to-possible-filename assumption is now broken.

sfnelson:
With Clojure 1.7 nearing release, this issue should be a priority. Reader conditional support is a huge boon for full stack clj/cljs developers, so it would be good to know that leiningen will be ready before 1.7 lands.

technomancy:
Issues become prioritized when someone chooses to work on them. The best way to make sure something happens is to make it happen.

sfnelson:
https://github.com/technomancy/leiningen/pull/1911

(also https://github.com/Raynes/bultitude/pull/28)

technomancy:
Nice! Could we get some clojurescript users to try this out and leave feedback?

sfnelson:
@technomancy: there is one function that I have not updated to support cljc that probably needs to be updated, but I don't understand its purpose. I was hoping you could explain what `source-in-project?` (https://github.com/technomancy/leiningen/blob/master/src/leiningen/compile.clj#L68) is intended to do?

It's only use is in the following function, `class-in-project?`, where it is called on the *parent* of a class file. The function itself adds `".clj"` to the parent and looks for the result in source dirs:

```
=> (class-in-project? project /path/to/target/classes/pkg/file.class)
=> (source-in-project? /path/to/target/classes/pkg compile-path source-path)
=> (.exists? #file /path/to/src/pkg.clj)
```

Is there a clojure compilation pattern that results in class files created in a subdirectory named after the compiling namespace?

sfnelson:
[...]

PR #1920, opened 2015-06-19
Find .cljc test files

puredanger:


technomancy:
It looks like this might be covered in #1827; can you take a look at that and see if it does what you need?

technomancy:
Please take a look at #1827 (on master) to make sure it covers what you intended to do with this pull request.

PR #1930, opened 2015-07-09
If jar exists, :auto-clean true, don't rebuild during uberjar

### LINKED ISSUE #1816: uberjar should not rebuild the jar with :auto-clean false and no changes
Follow up for irc conversation.

With lein 2.5.1,
- run `lein new app testapp`
- add `:auto-clean false` to `project.clj` (I also tried adding it to `uberjar` profile`)
- run `lein uberjar` twice
- Compilation runs only once, but jars are regenerated every time

This is very annoying in large-ish projects, especially when you need to run `uberjar` several times in a row (i.e. when building packages).

arrdem:
Fixes #1816 and thus makes users responsible for cleaning when appropriate.

### DIFF
--- a/leiningen-core/src/leiningen/core/user.clj
+++ b/leiningen-core/src/leiningen/core/user.clj
@@ -92,12 +92,12 @@
   []
   (or (getenv "LEIN_GPG") "gpg"))
 
-(defn- get-english-env []
-  "Returns environment variables as a map with clojure keywords and LANGUAGE set to 'en'"
-  (let [env (System/getenv)
-        keywords (map #(keyword %) (keys env))]
-    (merge (zipmap keywords (vals env))
-           {:LANGUAGE "en"})))
+(defn- get-english-env
+  "Returns env vars as a map with clojure keywords and LANGUAGE set to 'en'"
+  []
+  (let [env (System/getenv)]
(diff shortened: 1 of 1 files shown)


technomancy:
In the current codebase, `:auto-clean` defaults to true, but here it defaults to false.

arrdem:
Oops my bad. Fixing.

technomancy:
So, a related problem is that the point of `:auto-clean` is to make the defaults safe, but allow the user to disable it if they want speed over safety. It seems like this does the opposite--if `:auto-clean` isn't set, you'll get stale stuff in your uberjar. Is this accurate, or am I misreading it?

Really what we're talking about here is a cache expiry problem, isn't it? You're introducing a cache that never expires on its own. If we're going to add this, it needs to be done in a way that invalidates the jar file when the inputs change.

arrdem:
Yep. The system I'm working with and in the context of which this is interesting does lazy builds on the basis of a source tree hash and then does a `lein do clean, install, uberjar` for various reasons. I'm not sure under what conditions you want to expire the cached jar.

technomancy:
If you want this to happen by default, then it needs to expire the cached jar when the inputs to the jar change: in this case `project.clj` or the source files. If you are calculating the source changes in a separate system, then it needs to be opt-in.

PR #1931, opened 2015-07-10
Don't check certificate with wget in lein.bat

tommyettinger:
This is a potential installation blocker for new users on Windows. I
don't know how widespread the issue actually is because it may only be
in regard to older leiningen versions updating on Windows.


tommyettinger:
This whole commit may be a bad idea, but even though I don't know why the GitHub certificate is being rejected, it does block installation for me without this change.

technomancy:
Yeah, this is basically akin to jamming something in your seatbelt receptacle in order to make it stop beeping at you when you're driving with no seatbelt. Need to fix the underlying cause, not the symptoms.

PR #1933, opened 2015-07-14
Modify .lein-failures. Help retest individual tests.

shalu214patil:
Sometimes it is preferable and time-saving to retest :only failed tests,
instead of retesting the entire namespace.

To enable this, we can write a map to .lein-failures like this:
{"failed-ns-1" ["failed-test-1" "failed-test-2"]
 "failed-ns-2" ["failed-test-1"]
 ... ... ...}

But we do NOT change the API or behaviour of `lein-retest`.

We only modify it to read namespace names from a map instead of a set.

We now have the option to write a custom plugin or wrapper that retests
only failed tests.

### DIFF
--- a/src/leiningen/retest.clj
+++ b/src/leiningen/retest.clj
@@ -10,6 +10,6 @@
     (apply test/test project
            (concat (if (.exists (java.io.File. ".lein-failures"))
                      (->> (slurp ".lein-failures")
-                          read-string sort (map name)))
+                          read-string keys sort))
                    selectors))
     (main/abort "Cannot retest when :monkeypatch-clojure-test is disabled.")))
--- a/src/leiningen/test.clj
+++ b/src/leiningen/test.clj
@@ -71,7 +71,7 @@
        `(let [~ns-sym ~(form-for-select-namespaces namespaces selectors)]
           (when (seq ~ns-sym)
             (apply require :reload ~ns-sym))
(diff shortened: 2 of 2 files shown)


hyPiRion:
Hi there! Sorry for the slightly long response time here.

This looks good! I have just one question – currently, the emitted files look like this:

```clj
{"failed-ns-1" ("failed-test-2" "failed-test-1")
 "failed-ns-2" ("failed-test-1")}
```

Where `failed-ns-1/failed-test-1` is ran before `failed-ns-1/failed-test-1`. Do you think it is important to preserve the ordering? If so, then perhaps changing `(update-in @failures# [ns-name#] conj test-name#)` to `(update-in @failures# [ns-name#] (fnil conj []) test-name#)` would be an idea, to make it easier for retest plugins.

If it's not important to preserve ordering, then it might be an idea to store the values in a set instead (`(fnil conj #{})`). Then it will be more evident that there is no ordering.

Also, do you think you can remove the commit ed4aadc3917c234781558b15ac7ba1ce5373af28 from this pull request?

shalu214patil:
Updated test-names list to vector to maintain order. Also, removed commit https://github.com/technomancy/leiningen/commit/ed4aadc3917c234781558b15ac7ba1ce5373af28.

hyPiRion:
Works like a charm, thanks!

PR #1937, opened 2015-07-16
Bump cheshire dependency.

### LINKED ISSUE #1875: lein overwrites jackson dependency
1. create new project
2. add a dependency to jackson > 2.2 e.g. `[com.fasterxml.jackson.core/jackson-core "2.3.1"]`
3. `lein repl`
4. `user=> (clojure.java.io/resource "com/fasterxml/jackson/core/JsonFactory.class")`

Actual: the output shows that `JsonFactory` from the Leiningen jar (jackson 2.2, transitively added via cheshire) is used: 
`#<URL jar:file:/Users/tbayer/.lein/self-installs/leiningen-2.5.1-standalone.jar!/com/fasterxml/jackson/core/JsonFactory.class>`

Expected: The class from the 2.3.1 dependency is used. Otherwise e.g. the method `requiresPropertyOrdering` cannot be used, as it is `since 2.3`.

conormcd:
The current version pulls in really old versions of jackson which can easily break plugin code.

This doesn't fix #1875, but it might reduce the number of people adversely affected by this dependency.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -23,7 +23,7 @@
                  [clojure-complete "0.2.3"]
                  ;; bump versions of various common transitive deps
                  [slingshot "0.10.3"]
-                 [cheshire "5.0.2"]
+                 [cheshire "5.5.0"]
                  [clj-http "0.9.2" :exclusions [crouton]]]
   ;; checkout-deps don't work with :eval-in :leiningen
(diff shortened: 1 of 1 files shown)


technomancy:
Any reason not to use 5.5.0?

conormcd:
Good point, no. Fixed.

technomancy:
Cool; thanks.

PR #1941, opened 2015-07-23
added fix for components.xml merging.  If the description tag contain…

towler73:
…ed embedded html, it needs to be escaped in the resulting merged components.xml.


technomancy:
I think we can probably do better than hard-coding a list of characters to escape. Surely in that huge pile of libs we pull in this functionality exists already?

towler73:
That's a good point.  Let me look into that.

towler73:
I changed escaping to use commons.lang.StringEscapeUtils.escapeXml().

technomancy:
That looks better; thanks.

PR #1943, opened 2015-08-01
Update docs - SCP deploy to clojars not supported

danielcompton:
Relates to https://github.com/ato/clojars-web/issues/308


technomancy:
Good catch; thanks.

PR #1954, opened 2015-08-14
Allow check for root user to process correctly on Solaris.

orderthruchaos:
On Solaris, the standard `id` command, `/usr/bin/id`, does not allow for the `-u` flag.  `/usr/xpg4/bin/id -u` must be used instead.


hyPiRion:
Looks good, thanks!

Issue #1956, opened 2015-08-15
pprint runtime info as edn


laurentpetit:
Counterclockwise needs to synchronize the Eclipse project classpath with leiningen so that dependencies, source dependencies, native dependencies are correctly set.

At least for 2 purposes:
- leveraging as much as possible from the IDE in terms of static code introspection (suggesting java / javadoc from the listed dependencies and dependencies source jars)
- easy / quick startup of a REPL with all the dependencies (including native dependencies) without involving leiningen (works well in 80% cases and is the preferred way by many because startup time is faster)

Currently, what has been suggesting for "calling leiningen and get runtime information" is the following ( http://librelist.com/browser//leiningen/2015/8/14/getting-leiningen-classpath/ ):

```
lein update-in : assoc :eval-in :pprint -- run -m clojure.main
```

This works, but has at least 2 problems:
- Problem 1: what is output is not readable as edn. The first part looks like
      `
  Java: java
[...]

PR #1957, opened 2015-08-15
Add :pprint-edn to eval-in

laurentpetit:
- factorized code for runtime-info to emit intermediate datastructure
- use new runtime-info fn from eval-in :pprint (no regression)
- add new flavor eval-in :pprint-edn which pprints everything read-ably

### DIFF
--- a/leiningen-core/src/leiningen/core/eval.clj
+++ b/leiningen-core/src/leiningen/core/eval.clj
@@ -313,17 +313,33 @@
     (System/setProperty k v))
   (eval form))
 
-(defmethod eval-in :pprint [project form]
+(defn runtime-info
+  "Function returning command line info as pure data. Can be called by third party libraries"
+  [project]
+  {:java      (or (:java-cmd project) (System/getenv "JAVA_CMD") "java")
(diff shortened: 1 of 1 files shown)


laurentpetit:
First trial at a proposal for #1956

hyPiRion:
@laurentpetit Terribly sorry for the late response time here. This seems fine by me, but I wonder about one thing: Would it be easier or harder for you to read the output if we were to incorporate the form into the map? Currently it emits two values, which may be slightly awkward to work with.

I have no preference over either, but I would assume embedding the form into the map would be easier to work with for consumers.

technomancy:
Just took a look at this, and it looks great. Unless @laurentpetit wants to make the change to a single form containing both pieces of data, I'd like to go ahead and merge it.

laurentpetit:
Actually, it's so old that I doubt it will be used by Counterclockwise soon. Indeed the project is in maintenance mode, and I'm not working on it anymore.

Anyway, there's still value in the PR so feel free to merge it! Just saying!

technomancy:
OK, thanks.

Issue #1944, opened 2015-08-03
Compilation errors lost when compiling with :gen-class


brunchboy:
I have wasted many hours over the last couple weeks while working on [afterglow-max](https://github.com/brunchboy/afterglow-max) due to the fact that when I am compiling a Clojure source file which extends a Java class, any little mistake I make while coding gives me the following opaque output when I try to compile, rather than the helpful message telling me where the problem occurred that I am used to:

```
lein uberjar
Compiling afterglow.max.core
Compiling afterglow.max.Cue
Compiling afterglow.max.Eval
Compiling afterglow.max.init
Compiling afterglow.max.Metro
Exception: java.lang.UnsatisfiedLinkError thrown from the UncaughtExceptionHandler in thread "main"

Compilation failed: Subprocess failed
Uberjar aborting because jar failed: Compilation failed: Subprocess failed
Error encountered performing task 'uberjar' with profile(s): 'base,system,user,provided,dev,update'
Uberjar aborting because jar failed: Compilation failed: Subprocess failed
```

Then I would have to go through a process of tedious trial-and error to hunt down my mistake, since the actual compiler output is suppressed, I cannot find it anywhere. After discussing this a bit on the Clojure mailing list, I realized that I can be much more productive when this happens by directly invoking the Clojure compiler from the REPL, which does produce the sought-after compilation error details:

```
afterglow.max.core=> (compile 'afterglow.max.Metro)

CompilerException java.lang.RuntimeException: Unable to resolve symbol: set-bar in this context, compiling:(afterglow/max/Metro.clj:121:7) 
```

So I can limp along like this much better than I was before, but it would be nice if the compilation errors were displayed by Leiningen in this situation, like they normally are.

To reproduce this problem, you can clone the afterglow-max repo at https://github.com/brunchboy/afterglow-max, then comment out the `set-bar` function in `afterglow.max.Metro` and run `lein uberjar`.

hyPiRion:
Fixed by #1984.

brunchboy:
Nice! And the very next day after you helped me (as DeepSymmetry) with another issue on IRC. Quite the good week for Leiningen support for me. :grin:

PR #1962, opened 2015-08-21
Issue #1944

shimanekb:
Solved issue #1944.

On line 142 the (.printStackTrace t#), t# being the throwable, was not printing. I fixed this by changing line 142 using println and calling .getMessage on t#.


hyPiRion:
Could you wrap the println in a `(binding [*out* *err*] ...)` in the same way we emit "Compiling ..." messages? `.printStackTrace` prints on stderr by default, so it'd be nice to keep the error messages there.

shimanekb:
sure

PR #1972, opened 2015-09-02
Feature/honor sh return codes

### LINKED ISSUE #1968: If git tag fails, lein silently ignores
Is this intentional behavior? Or could it be fixed/a pr made to the effect that if the tag fails to be created, the build fails.

Should be as simple as checking the return code of this call, I think. https://github.com/technomancy/leiningen/blob/9c5192e6520fb96aa625313a468e0143001f54ca/src/leiningen/vcs.clj#L67

timvisher:
Still have some functional tests to do but this seems reasonable. I'm not super thrilled about the name `sh-with-exit-code` but I couldn't think of a better one off the top of my head.

Closes #1968 

Let me know if I'm not on the right track, otherwise I'll let you know when I'm done functionally testing it.

- [x] Add unit tests
- [x] Run a successful release
- [x] Run a failed release (gpg tag failure specifically)

### DIFF
--- a/leiningen-core/src/leiningen/core/eval.clj
+++ b/leiningen-core/src/leiningen/core/eval.clj
@@ -185,6 +185,21 @@
             (.resurrect System/in))
           exit-value)))))
 
+(defn sh-with-exit-code
+  "Applies SH to CMD and on a non-0 exit code, throws an Exception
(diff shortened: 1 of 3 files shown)


timvisher:
Good to go.

hyPiRion:
I tested this on a broken git repo, and it works as expected. Thanks!

PR #1974, opened 2015-09-04
Create source, resource and test paths for eval prep

weavejester:
If a directory on the classpath does not exist when the JVM is started, it's ignored even if it is subsequently created. This can be a pain when dealing with resources generated on the fly, such as Javascript generated from a Clojurescript source. Currently the only solution is to close and restart the REPL when that happens.

This pull request ensures that source, resource and test paths exist as part of the prep work for eval-in-project, thereby fixing this issue. I've added a test that fails without these changes.


hyPiRion:
Yeah, this seems like a straightforward fix for a painful issue. Thanks!

Issue #1969, opened 2015-08-28
leiningen command-line update-in has lower priority than ~/profiles.clj


laurentpetit:
CCW, when launching is project server REPL, is forcing a project's JVM to start in debug mode by adding this to the leiningen command line: `lein .... update-in :jvm-opts concat ["-Xdebug" "-Xrunjdwp:transport=dt_socket,server=y,suspend=n"] -- repl`.

Many people seem to have `{:user {:jvm-opts ^:replace []}}` in their `profiles.clj`. When it is so, then the `update-in ...` command-line stuff is silently ignored as can be seen by inspecting the project's process command line created by Leiningen.

So, a whole category of people are currently unable to launch their project with a REPL in Debugging mode.

Is it normal that what is declared on the command line does not take precedence over what is declared in `~/.lein/profiles.clj`(or even `./project.clj` - not tested`) ? I would say it's not normal.

What do you think? Bug? Or should CCW be doing things differently to force the insertion of the remote JVM debug options to the project process through invoking the leiningen command?

technomancy:
No, this is definitely a bug. The profile setting should not be able to override what's explicitly entered in the update-in task.

laurentpetit:
@technomancy thanks for the quick qualif!
Do you thing this is a low hanging fruit I could try to provide a PR for, or is it a delicate part of the code that should better be handled by leiningen internal experts? If low hanging fruit, can you provide some starter point?

technomancy:
How easy this is depends on whether it's a bug in the `update-in` task or deeper in the merge logic. My hope is that it's a bug in the task which should be fixed by a fairly straightforward patch there; if it's deeper inside `leiningen-core` then it'll take some more digging to uncover.

hyPiRion:
As a workaround and FYI, `{:user {:jvm-opts []}}` will replace the default `:jvm-opts` with `[]` and should work fine with the CLI call you present. I guess many people use `{:user {:jvm-opts ^:replace []}}` to ensure that default optimisations are set, but the `^:replace` part is not needed anymore.

hyPiRion:
Also, I've looked into this due to #1981, and it seems that the problem at hand here is quite convoluted because I'm not entirely sure whether this _is_ expected functionality or not.

Many tasks merge in extra profiles when they need or where it's expected – perhaps the best known example is uberjar, but repl also does this. In a sense, you can consider those tasks of expanding into something equivalent of `lein with-profile +some,profiles actual-task`.

[...]

PR #1977, opened 2015-09-07
improve project map merge logic

j-allard:
This should fix this https://github.com/technomancy/leiningen/issues/1969
My understanding of the merge process isn't all that clear so I'm open to any kind of feedback.

Thanks :+1: 

EDIT: Okay this definitely need more work, sorry for the noise. I'll reopen when its ready.


j-allard:
Okay it obviously broke

PR #1980, opened 2015-09-08
Make the file-not-found test less strict

j-allard:
The JVM sometimes prepends stuff to the err output (like if you have the JAVA_TOOL_OPTIONS set) and this makes this particular test fail.
This adds a bit of convenience and this doesn't invalidate the test AFAIK.


hyPiRion:
Yeah, this shouldn't cause any trouble. Thanks.

PR #1981, opened 2015-09-08
Remove metadata on updated key with update-in

j-allard:
Okay so here my second take on this.
Sorry for the "duplicate" PR, I couldn't reopen the other one because I forced-pushed to my branch :(

FIxing this https://github.com/technomancy/leiningen/issues/1969
Continuity of this https://github.com/technomancy/leiningen/pull/1977

So basically this is removing the existent metadata on keys you update with the update-in task and this makes the rest of the merge process to behave as expected (i.e the keys updated using update-in have priority).

Any thoughts or something obvious I might have missed ?

### DIFF
--- a/src/leiningen/update_in.clj
+++ b/src/leiningen/update_in.clj
@@ -13,13 +13,20 @@
      (mapv read-string f-args)
      task+args]))
 
+(defn discard-metadata [project keys-vec]
+  (if-not (nil? (get-in (meta project) keys-vec))
+    (vary-meta project clj/update-in keys-vec #(with-meta % {}))
+    project))
+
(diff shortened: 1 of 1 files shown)


hyPiRion:
Unfortunately, this doesn't solve all cases of this problem: If you only drop metadata, then merging other profiles in would merge "as usual". E.g. if you do `lein update-in :jvm-opts concat ["-Xdebug"] -- repl`, then if the :repl profile contains `:jvm-opts ^:replace []` you'll end up with an empty vector for :jvm-opts. The question, of course, is whether this is "as expected" or not – I've explained this issue in #1969.

I think I want to defer fixing #1969 until 3.0.0 for now, unless a more immediate use case appears.

PR #1984, opened 2015-09-09
issue #1944 Fix

shimanekb:
Fixed issue #1944 using std out and err, as suggested.


hyPiRion:
Whoops, this seems to just bind `*out*` to `*err*`, not including the `(println  (.getMessage t#))` change. Do you think you could replace `(.printStackTrace t#)` with `(println  (.getMessage t#))` in this PR?

You can just add a new commit to this PR, it's not super important to have it in a single commit.

shimanekb:
Sure

shimanekb:
ok, I made the changes.

hyPiRion:
Excellent, thanks!

PR #1986, opened 2015-09-10
Add option to not sign tags when running "vcs" "tag"

### LINKED ISSUE #1799: Option to not sign git tags
Many projects like to sign tags (including this one itself). Many more others do not.

There has been previous discussion (http://www.raynes.me/logs/irc.freenode.net/leiningen/2014-09-03.txt) as to why the leiningen source code repository has signed tags, and those points are undisputed.

Other projects do not require tag signing. Even open-source projects with broad contribution may not desire tag signing and use a different mechanism to "sign off" contributions (cf. http://mikegerwitz.com/papers/git-horror-story).

Whatever the case, the user should have the option to either sign or not sign tags. Given the leiningen team sign by default, it may be the option is to turn off signing (even though that is the inverse of git defaults). It would be preferred if the option was set using the existing 'profiles' mechanisms (e.g. `:vcs-options {:tag-sign false}`). Preferably, the `lein vcs tag` should also take a flag to turn off signing that overrides the profile merging (e.g. `lein vcs tag --unsigned|-u`.

Given the limited degrees of freedom of `lein vcs tag` implies so-called "lightweight tags" in git (that is, `git tag` with no options).

### LINKED ISSUE #1873: lein release can't create tag without gpg signature.
It would be useful to have an option that turns off the `-s` option.

danielcompton:
- Add vcs tag commandline argument parser
- Update documentation to show new options

This is my first non-doc contribution to Leiningen, so let me know if any of my code style is wrong. I'd also welcome any improvements on the docs for tag arguments, I felt my changes were a bit awkward.

Fixes #1873, fixes #1799

- [ ] Need to notify https://github.com/nberger/lein-tag-no-sign about this change.

### DIFF
--- a/doc/DEPLOY.md
+++ b/doc/DEPLOY.md
@@ -297,6 +297,13 @@ libraries using Leiningen. Applications will have different requirements
 that are varied enough that Leiningen doesn't attempt to support them
 out of the box.
 
+### Tagging
+
+By default `["vcs" "tag"]` will create a GPG signed tag with your project version
+number. You can add a tag prefix by passing the prefix after `"tag"`,
+for example: `["vcs" "tag" "v"]`. You can disable tag signing by passing `--no-sign`,
+for example: `["vcs" "tag" "v" "--no-sign"]` or `["vcs" "tag" "--no-sign"]`.
+
 ## Deploying to Maven Central
 
 Deploying your libraries and other artifacts to [Maven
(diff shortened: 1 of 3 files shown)


danielcompton:
I'm also happy to change/add the arguments provided for indicating sign/nosign.

hyPiRion:
There is only one super small thing I'd like to address: Isn't it more common to use `--no-foo` instead of `--nofoo`? clojure/tools.cli and git for example follows this rule. For consistency, I think it makes sense to follow this convention.

Apart from that, the patch looks excellent and I'd happily merge it in.

danielcompton:
I've also verified that this actually works using the `lein vcs tag --no-sign`.

hyPiRion:
great, thanks! :)

PR #1990, opened 2015-09-21
allow setting no-auth in settings

MichaelBlume:
Some wagons can find their own authentication. For example, the S3 wagon can
use IAM roles if it's running on an EC2 instance. For these wagons, all we need
to do is get out of the way.


hyPiRion:
Looks good, thanks!

PR #1993, opened 2015-09-23
Added to template docs explaining mustache delimiters

rwilson:
I ran into this, so I added a sub-section to the templates doc with an example of how to avoid mustache tag delimiters conflicting with clojure syntax. The example is contrived, but it demonstrates the problem succinctly.


hyPiRion:
Thanks; looks good. :)

PR #1998, opened 2015-09-28
Don't trigger implicit-aot-warning on regex match

emlyn:
Prevent the implicit-aot-warning from triggering if `:aot` contains a regex matching the main namespace, e.g. if `project.clj` contains something like:
```clojure
:main project.core
:aot [#".*"]
```
I'm not sure if this is the best way to fix this, and I haven't actually tried it out, so consider this more of a bug report than the actual fix.

### DIFF
--- a/src/leiningen/compile.clj
+++ b/src/leiningen/compile.clj
@@ -7,7 +7,7 @@
   (:refer-clojure :exclude [compile])
   (:import (java.io PushbackReader File)))
 
-(defn- regex? [str-or-re]
+(defn regex? [str-or-re]
   (instance? java.util.regex.Pattern str-or-re))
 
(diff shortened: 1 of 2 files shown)


hyPiRion:
Hi there, thanks for the report and the pull request!

This looks almost done, and I think the implementation idea is sound. However, there's a tiny bug that has to be fixed before it's merged into master: `#(re-matches % (:main project))` has to be replaced with `#(re-matches % (str (:main project)))` – otherwise we will get a complaint from re-matches that it cannot convert a Symbol into a CharSequence.

If you add a commit that fixes that I'll be happy to merge it in :)

emlyn:
Thanks for the response, I've made that change :-)

hyPiRion:
Excellent, thanks!

PR #2004, opened 2015-10-11
Return non zero status when :pedantic? is set to true or :abort

### LINKED ISSUE #1856: Return non zero status when :pedantic? is set to true
Could a non zero status be returned when pedantic is enabled  (:pedantic? true in project.clj) and lein deps :tree is executed when there are conflicting dep versions? This would make it much easier to integrate into a CI tool such as teamcity. Currently a zero status is returned what ever the result.

pietromenna:
Closes #1856.


hyPiRion:
Looks good, thanks! :)

PR #2006, opened 2015-10-14
Documentation fix: change passphrase to password

RasterBurn:
`:passphrase` does not work in this context (non-gpg encrypted credentials).  This works when you specify `:password` instead.


hyPiRion:
Thanks!

Issue #1854, opened 2015-03-12
lein run messes with shutdown hooks


amalloy:
I'm not sure if this is expected behavior or what, but lein did something surprising when there were JVM shutdown hooks present. When running my project via `lein run -m whatever`, ctrl-c seems to forcibly kill the process, without waiting for shutdown hooks to finish; `lein trampoline run -m whatever` doesn't have this behavior, instead allowing all shutdown hooks to run to completion.

I tried adding `:eval-in :trampoline` to my project, since a reading of leiningen.core.eval suggests that is a thing I might be able to do, but this just makes `lein run -m whatever` exit immediately without running my code at all, so I gather that is not something I'm supposed to do.

hyPiRion:
Yeah, that happens because we forcibly kill subprocesses:

https://github.com/technomancy/leiningen/blob/14379a7305cb9d99eab9f63fba0ee8769b90280c/leiningen-core/src/leiningen/core/eval.clj#L168-L169 – seems reasonable to do something else than `.destroy()` if it's possible.

Happy to take a patch which handles this.

jhalterman:
FWIW, this applies to `lein test` as well.

evanostroski:
Ok, so I'm a bit of a newbie, not just to this project, but to FOSS period. Are you wanting me to modify this pull to handle this other bug as well?

hyPiRion:
@evanostroski: Oh, sorry – I don't usually leave comments while I'm in the process of reviewing a pull request. I should probably reconsider that.

evanostroski:
@hyPiRion No worries, I was just getting curious!

hyPiRion:
I tried to repeat this issue, but it seems like this is Windows/cygwin-specific. Sending a sigterm or sigint to a java process on OS X/Linux seems to propagate that term/int message downstream. Which seems to make sense, given that Ctrl-D and Ctrl-C works as expected in a repl on those systems.

evanostroski:
Okay, I can install cygwin and run some tests. I'm guessing the print3 app you wrote would be a good start?

jhalterman:
I was actually hitting this on OSX - my shutdown hook was never called when doing ctrl-c.

hyPiRion:
Ugh, this is slightly more bothersome than I initially thought (hence the removal of Newbie).

[...]

PR #2015, opened 2015-10-31
allow subprocesses to shutdown gracefully

evanostroski:
addresses 'lein run messes with shutdown hooks #1854'

solution based on this thread in core-libs-dev mailing list:
http://mail.openjdk.java.net/pipermail/core-libs-dev/2014-March/025525.html

Instead of terminating the subprocess outright with Process.destroy, this
patch opens a outputstream to the subprocess stdin and sends Ctrl-C to end
it.  This allows the subprocess to call its own shutdown hooks.

A with-open is NOT used here because after the process terminates, there's
nothing to close and it throws IOException Stream Closed.

### DIFF
--- a/leiningen-core/src/leiningen/core/eval.clj
+++ b/leiningen-core/src/leiningen/core/eval.clj
@@ -170,7 +170,8 @@
   (let [env (overridden-env *env*)
         proc (.exec (Runtime/getRuntime) (into-array String cmd) env (io/file *dir*))]
     (.addShutdownHook (Runtime/getRuntime)
-                      (Thread. (fn [] (.destroy proc))))
+                      (Thread. (fn [] (let [in (.getOutputStream proc)]
+                                  (.write in 3)))))
     (with-open [out (.getInputStream proc)
                 err (.getErrorStream proc)
                 in (.getOutputStream proc)]


hyPiRion:
I'm on this, but I need to understand what's going on in my system first. I initially thought that sending just an arbitrary 0x03 to the process (as with e.g. `echo -n -e '\0x03' | hd`) would just send that specific byte. However, I can't find any signs of said value, which makes me confused.

hyPiRion:
I wanted to ensure that this doesn't send off a random 3 into the subprocess, so I added in a flush and some sleeping to the shutdown hook as follows:

```clj
(.addShutdownHook (Runtime/getRuntime)
                  (Thread. (fn [] (let [in (.getOutputStream proc)]
                                    (.write in 3)
                                    (.flush in)
                                    (Thread/sleep 2000)))))
```

Then, I made a Clojure app by calling `lein new app print3` and edited the core.clj so that it contained the following:

```clj
(ns print3.core
  (:gen-class))

(defn -main
  [& args]
  (loop [v (.read System/in)]
    (prn v)
    (if-not (= v -1)
      (recur (.read System/in)))))
```

At least on Linux, when I run `lein-master run` and then `kill LEIN_PID`, I will see 3 printed out. However, the subprocess will not die.

The way it works now with `.destroy`, at least on Linux, a sigterm is issued downstream: If you add `[beckon "0.1.1"]` as a dependency and then fixup the file like so:

```clj
(ns print3.core
  (:require beckon)
  (:gen-class))

(defn -main
  [& args]
  (let [int-func (fn [] (spit "sigint" "Received INT signal, ignoring it"))
        term-func (fn [] (spit "sigterm" "Received TERM signal")
                    (System/exit 0))]
    (reset! (beckon/signal-atom "INT") #{int-func})
    (reset! (beckon/signal-atom "TERM") #{term-func}))
  (loop [v (.read System/in)]
    (prn v)
    (if-not (= v -1)
      (recur (.read System/in))))
  (Thread/sleep 100000))
```

You will see a file named sigterm in the directory you issued `lein run` at, at least on Linux. (And you will not see any sigint)

Does this behave differently on Windows/Cygwin? I would be surprised if the 3 is not printed out, but I would expect the sigterm/sigint files to not be there on some platforms (which is the cause of #1854 in the first place).

hyPiRion:
Also, for what it's worth, it's probably nice to know that the questions mentioned in the link you provided:

> Are you in control of the code that represents the child sub-process? Is this code already using System.in (STDIN) stream for anything? If the answers are YES and NO respectively, then you may be able to use the System.in in the sub-process

Are NO and MAYBE, respectively.

hyPiRion:
After some more digging, it seems like this seems to be impossible to do on the JVM correctly on all systems.

Sorry about labeling the issue as newbie, apparently that was totally not the case =|

evanostroski:
No problem, I learned from it, which is always a win in my book.  Thank you for your responses!

PR #2017, opened 2015-11-01
Bump the nREPL dep to 0.2.12

bbatsov:
nREPL 0.2.11 added an extremely important source-tracking feature, which is leveraged by CIDER (and probably some other tools). 0.2.12 fixed a critical bug introduced in 0.2.11.


hyPiRion:
Thanks! I also bumped some hidden nREPL deps that aren't that easy to know about.

bbatsov:
Thanks!

Issue #2019, opened 2015-11-04
Support Windows "Take Command" alternative shell in lein.bat


m9aertner:
I'm using an alternative Windows Shell named "Take Command", not plain CMD.EXE. See [https://jpsoft.com/] (the TCC/LE variant can be downloaded/used free of charge).

I find that two minor amendments are required to make this work _cleanly_ with `lein.bat`
1. One delete command does not redirect stdout, unlike other similar existing uses in `lein.bat`.
2. The classpath does not get set in TCC, simply re-ordering the condition helps.

I'll provide a pull request. Note that these amendments are compatible with CMD.EXE.

hyPiRion:
Hi there,

Thanks for looking into this – I would obviously have no issues with a pull request fixing this :)

m9aertner:
Let me do some more testing, by following the CONTRIBUTING.md instructions...

hyPiRion:
Just a heads up, #1899 tells that there may be some issues building the uberjar on Windows – that may also be an issue for testing lein itself. Since this is a modification of `lein.bat` though, you could try it out on a copy of the `lein.bat` you use in day-to-day work.

m9aertner:
Solved as part of #1899, see PR #2023

PR #2020, opened 2015-11-04
Support Windows "Take Command" alternative shell in lein.bat #2019

m9aertner:


hyPiRion:
Seems good, just one question: You mentioned in #2019 that delete commands should redirect to stdout. Is this also the case for the line [367 in lein.bat](https://github.com/m9aertner/leiningen/blob/050f6308bf09e3a2e751baf768bbdc0f0e1fc360/bin/lein.bat#L367)? We could maybe also fix this one while we're at it.

m9aertner:
Yes, you're probably right. TCC shows "1 file deleted" on **stdout** (different from CMD) there. I had only seen/addressed the "0 file deleted" message on the delete above. Forget the PR, it's not complete, clean bootstrapping is also not yet working perfectly.

Issue #2021, opened 2015-11-07
Improvement: Support binary file generation in lein-new


sbondaryev:
Currently we can't generate binary files (images etc.) with 'renderer' function.
So lots of clojure web-projects use their own implementation like:
```clj
(defn resource [r]
(->> r (str "leiningen/new/cryogen/") (io/resource) (io/input-stream)))
```
Could it be useful to add this function in templates.clj?

hyPiRion:
Yeah, that wouldn't be a bad addition. Happy to take a PR for that.

sbondaryev:
I replaced default render io/reader by io/input-stream
is it a right solution?

hyPiRion:
Fixed by #2022.

PR #2022, opened 2015-11-07
support binary files in lein-new

sbondaryev:
replace io/reader by io/input-stream to support binary files
issue https://github.com/technomancy/leiningen/issues/2021

### DIFF
--- a/src/leiningen/new/templates.clj
+++ b/src/leiningen/new/templates.clj
@@ -141,8 +141,20 @@
         (if-let [resource (io/resource path)]
           (if data
             (render (slurp-resource resource) data)
-            (io/input-stream resource))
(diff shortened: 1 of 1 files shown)


hyPiRion:
I would prefer writing binary files to be a new function you'd have to call explicitly. The reader exposed right now may be used by templates, and I'd like to not break compatibility with those right now.

Would it be an idea to make a new function with the name `raw-resource` or something similar?

sbondaryev:
I've moved functionality to raw-resourcer which returns the function to generate binary files

hyPiRion:
Thanks! I'll add in a mental note about moving them into the same function for 3.0.0.

PR #2009, opened 2015-10-23
show a better error message when java is not found in PATH

propan:

PR #2024, opened 2015-11-09
Don't preset JAVA_CMD. Refs #2009

juergenhoetzel:
Because on Cygwin JAVA_CMD will be set to an UNIX path. This will result
in a "File not found" exception when the native Leiningen JAVA process
will execute JAVA_CMD.

Also use shell builtin "type" fs external "which"

### DIFF
--- a/bin/lein
+++ b/bin/lein
@@ -222,12 +222,7 @@ else # Not running from a checkout
     fi
 fi
 
-if [ -z "$JAVA_CMD" ]
-then
-    JAVA_CMD=$(which java)
-fi
-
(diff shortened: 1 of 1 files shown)


hyPiRion:
Thanks! :)

I also added in 374557c70af5cf94a6bbed3dc62bd13e4853cc25 into the pull request branch. This should set `LEIN_JAVA_CMD` to `JAVA_CMD` if set, else `java`. Some people use JAVA_CMD to specify which java command Leiningen uses, so we still need to support that.

PR #2026, opened 2015-11-15
Reset ERRORLEVEL and RC in lein.bat

### LINKED ISSUE #1865: lein.bat does not propagate execution errors
You need to save %errorlevel% after calling java.exe and then use it after EOF (e.g. exit /b %errorlevel%) so you propagate errors out of the script.

winks:
Fixed #1865
For reference:
http://blogs.msdn.com/b/oldnewthing/archive/2008/09/26/8965755.aspx


hyPiRion:
Any Windows users want to test this out? I'd like to get this into 2.6.0 and I feel pretty confident it's hard to mess this one up, but it would be preferable to have it tested beforehand.

hyPiRion:
Guessing this would be juust fine to merge in.

Issue #2035, opened 2015-12-04
Empty stack trace inconsistencies


solicode:
I often lose stack trace information unless I explicitly set `-XX:-OmitStackTraceInFastThrow`

I'm aware this option needs to be turned off to avoid those `[trace missing]` messages, but I was under the impression that Leiningen sets `-XX:-OmitStackTraceInFastThrow` by default (#1025). However, that doesn't appear to be the case. I still run into the problem with the default settings. And when I run:

```clojure
(.getInputArguments (java.lang.management.ManagementFactory/getRuntimeMXBean))
```

I can see that `-XX:-OmitStackTraceInFastThrow` is not there.

I took a look at the source code and [this part](https://github.com/technomancy/leiningen/blob/2.5.3/leiningen-core/src/leiningen/core/project.clj#L468-L471) seems to be the cause. I wanted to come up with a fix, but I'm actually not sure I understand the logic behind the "Tiered" check in order to attempt that.

I don't know what the intended behavior is supposed to be, but all that I know is the default behavior is inconsistent depending on how you installed Leiningen:

- If you use [bin/lein](https://github.com/technomancy/leiningen/blob/2.5.3/bin/lein), you get `-XX:-OmitStackTraceInFastThrow` by default because `LEIN_JVM_OPTS` includes the word "Tiered".
- If you used a package manager to install Leiningen (like I did), you probably got [bin/lein-pkg](https://github.com/technomancy/leiningen/blob/2.5.3/bin/lein-pkg). But in this file, `LEIN_JVM_OPTS` does not have "Tiered" in it (instead, it's passed separately).
- And if you're on Windows using [bin/lein.bat](https://github.com/technomancy/leiningen/blob/2.5.3/bin/lein.bat), "Tiered" doesn't even show up in that file.

Am I understanding this correctly? Like I said, I don't know what the "correct" behavior is supposed to be, but I can't imagine this inconsistency is intentional.

I think ideally `-XX:-OmitStackTraceInFastThrow` would be set in `:base` 100% of the time, and it would be up to the user to override that default behavior.

hyPiRion:
Thanks for tracking this one down.

Yeah, I think the behaviour right now is inconsistent. I also agree that we should pass in `-XX:-OmitStackTraceInFastThrow` by default, with options to override.

solicode:
Thanks. If that's the case, here's what I'm thinking:

[...]

PR #2037, opened 2015-12-05
Modify tiered-jvm-opts for :base profile

solicode:
Fix for #2035 to always prevent "trace missing" messages by default when exceptions occur.


hyPiRion:
Looks good, thanks!

PR #2044, opened 2015-12-18
After AOT compiling, shutdown agents for expedient exit

ryfow:
If compiling a namespace calls a function that uses Clojure's agents
feature and `shutdown-agents` is not called, the process `compile`
started will wait up to 60 seconds before exiting.

It's not clear to me how to write a good test case for this. I'm open to suggestions.

I did publish a very simple project that reproduces the stalling at https://github.com/ryfow/compile-stall

### DIFF
--- a/src/leiningen/compile.clj
+++ b/src/leiningen/compile.clj
@@ -132,16 +132,18 @@ Code that should run on startup belongs in a -main defn."
   ([project]
      (if-let [namespaces (seq (stale-namespaces project))]
        (let [ns-sym (gensym "namespace")
-             form `(doseq [~ns-sym '~namespaces]
-                     ~(if main/*info*
-                        `(binding [*out* *err*]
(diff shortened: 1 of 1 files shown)


ryfow:
That's odd. I ran the tests locally and didn't run into an issue. I'll look into the failures.

trptcolin:
One thing to watch out for here: since `shutdown-agents` is a global (and afaik irreversible) action, this may be at risk of breaking tools that want to embed Leiningen (e.g. Cursive Clojure was doing this at one point).

ryfow:
It appears to me that eval-in-project starts a new process for each call, so I don't think this would affect embedding. On the other hand, the travis-ci test failures indicate that agent requests can still be submitted after the code I inserted.

I'm closing this and created a new PR #2045 that takes an alternate approach.

PR #2045, opened 2015-12-18
Set short timeout on compiler agent thread pool

ryfow:
When namespaces use agents during load time, threads
are started in the agent threadpool which take 60 seconds to
end. This adds up to 60 seconds of stall time after the compile
task runs.

This patch checks to see if clojure.core/set-agent-send-off-executor!
exists, and if so, calls it with a Thread pool that has the keep
alive time set to 100 milliseconds.

Technically this could cause Thread churn, but practically most
namespace loading doesn't fire up enough Threads for thread
instantiation costs to add up to much.

I had a heck of a time trying to get rid of reflection warnings. I'd be interested in knowing what's going on.

### DIFF
--- a/src/leiningen/compile.clj
+++ b/src/leiningen/compile.clj
@@ -116,6 +116,22 @@
          (map #(if (string? %) (read-string %) %))
          (sort-by (comp not regex?)))))
 
+(def ^:private thread-factory-form
+  `(let [counter# (atom 0)]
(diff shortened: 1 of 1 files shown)


hyPiRion:
Yeah, this seems to be fine, thanks!

I would be surprised if this causes any issue during _compilation_ – as you say, very few compile invokations need agents in the first place.

Issue #1996, opened 2015-09-26
Add changes in `lein` to `lein-pkg`


hyPiRion:
We've had some changes in `lein` that haven't been moved over to `lein-pkg`.

TrangPham:
What are the changes that need to be moved over? 
Would this involve making the lein-pkg pass in `Tiered` in the `LEIN_JVM_OPTS`? https://github.com/technomancy/leiningen/blob/master/bin/lein-pkg#L107

hyPiRion:
Hi!

I have noted down the patch #1918 mostly, but yes, the `Tiered` stuff  in `LEIN_JVM_OPTS` would be good to have fixed too. There also seems to be a command line argument mismatch – lein-pkg has for example `-client -XX:+TieredCompilation` as arguments when calling `LEIN_JAVA_CMD`, and lein proper has an additional `-Dmaven.wagon.rto=10000`. I believe those arguments should be identical.

Happy to take patches for this, they don't have to fix all the differences between lein-pkg and lein.

TrangPham:
lein-pkg and lein seem to have a lot of similar code. Is there a reason why a common included bash script is not used? This would help keep it DRY.

hyPiRion:
Yeah, it's unfortunate that lein and lein-pkg doesn't share code. One of the issues is that people need to be able to download `lein` standalone from the git repo, whereas some downstream package managers (like Homebrew) uses `lein-pkg` directly from the repo. Consequently `lein` and `lein-pkg` must be standalone scripts.

I'd love some way to have the shared code in a single file and some script that regenerates `lein` and `lein-pkg`, but that may be harder to understand. Happy to take suggestions on how to make this better though, maybe it's not as bad as I expect it would be.

TrangPham:
**Option 1:**
One solution is to use bash to source a common file: http://stackoverflow.com/questions/192292/bash-how-best-to-include-other-scripts

Though this may not work if downstream package managers don't grab any of the other files.

**Option 2:**
A simple generation technique would to have a common file, a gen file for `lein` and a gen file for `lein-pkg`. The gen files would contain the unique code for `lein` and `lein-pkg` respectively. Then use a simple script that copies the common file over to `lein` and `lein-pkg` and appends the corresponding gen file to it.

ie.

    cp common lein
    cat gen_lein >> lein

**Option 3:**
[...]

PR #2047, opened 2015-12-19
Updating lein-pkg jvm opts to match lein

TrangPham:
Issue: #1996

I'm not sure if this is what you meant by the options needing to be the same.


hyPiRion:
Thanks

This is more or less everything, but do you think you could add in https://github.com/technomancy/leiningen/blob/a4496915e899a447965f14ed9a0f14634d7f383b/bin/lein#L153 as well? It should decrease startup time significantly.

TrangPham:
Okay added.

hyPiRion:
Excellent, thanks!

Issue #2043, opened 2015-12-17
Lein release does not respect version precedence


arichiardi:
Hello guys, it is the first time I use `lein release` so I might be doing something wrong.

I thought that I was on `0.1.3-SHAPHOT` and run `lein release :alpha` would bring me to `0.1.3-alpha`. At the of the day, SemVer says:

> Major, minor, and patch versions are always compared numerically. Example: 1.0.0 < 2.0.0 <
> 2.1.0 < 2.1.1. When major, minor, and patch are equal, a pre-release version has lower 
> precedence than a normal version. Example: 1.0.0-alpha < 1.0.0.

This did not happen and I directly deployed `0.1.3`.

I have not explored whether it is just `:alpha` the problem, but I can make some test with `:beta` and `:rc` as well.

hyPiRion:
Hi there,

I believe this is a documentation issue. I'm pretty sure this would've worked if you specified `0.1.3-alpha1-SNAPSHOT` -- in which case `lein release :alpha` would release 0.1.3-alpha1 and bump the version to `0.1.3-alpha2`.

arichiardi:
Yes but then how to generate `0.1.3-alpha-snapshot` with lein release. I can change the doc no problem, just to understand the process

arichiardi:
Not even with `-SNAPSHOT` works, I had `0.1.4-alpha1-SNAPSHOT` and after `lein release :alpha` I got:

```
* 0f98742 (HEAD -> master, tag: 0.1.4) Version 0.1.4
* b56f6d6 (origin/master) Version 0.1.4-alpha1-SNAPSHOT
```

hyPiRion:
Alright, then this seems to be a bug.

cprice404:
So, the expected behavior is that with an initial version of `0.1.4-alpha1-SNAPSHOT`, a `lein release :alpha` would result in an actual release of `0.1.4-alpha1` and then bump the version to `0.1.4-alpha2-SNAPSHOT`?

That is the behavior that I would expect, but the documentation doesn't really explain the intended behavior for the alpha/beta/rc stuff.

This is related to some of the ideas I'm playing with for #2049 and I think it's possible to make it work the way I described above, but I'm not sure if that would necessarily end up being backward-compatible with the existing behavior for alpha/beta/rc.

cprice404:
Here are some examples of how it works right now:

```
user=> (release/bump-version "1.0.0-SNAPSHOT" :alpha)
[...]

Issue #2049, opened 2015-12-22
Add ':build' version number to release task, to support 4-digit version numbers


cprice404:
Currently the 'change' part of the lein release task supports 3 version number levels: `:major`, `:minor`, and `:patch`.

Some of our projects use a 4th digit to represent the build version.

Would you be amenable to a PR that add support for a value of `:build` being passed in for the `change` level?

hyPiRion:
I am happy to take a patch for this if this is backwards compatible and well documented.

cprice404:
Awesome, thanks.

cprice404:
Started poking at this today and it seems like it would be easy to implement and write tests for; the one question / concern that I have is that the comments in the code repeatedly refer to the SemVer specification from http://semver.org, and that spec does not allow a 4th digit in the version number.  It seems to recommend putting a build number in the qualifier string after the hyphen, e.g. `1.0.0-25`.

Maven's version number scheme sounds like, while it will technically support 4-digit version numbers, it will probably screw up any tooling that attempts to do version number comparisons:

https://docs.oracle.com/middleware/1212/core/MAVEN/maven_version.htm#MAVEN400

Meanwhile, both RPM and DEB package versioning rules seem to support an arbitrary number of dot-separated numerics, and do not allow hyphens.  So, yay for standards and practices.  :(

We automatically build RPMs and DEBs from our projects, so conforming to their versioning rules is really what was inspiring this feature request, but it seems like it probably makes more sense to keep leiningen in compliance with the way that maven wants to version things... so I'm waffling on whether or not this is a good idea for an official leiningen patch or not.

cprice404:
I decided that it might make more sense to do the 4digit-version thing as a separate lein plugin.  I did my best to make it backward-compatible with the built-in release/change stuff.

Code is here:

https://github.com/puppetlabs/lein-release-4digit-version

Hoping to do a clojars release soon and then will probably add that to the plugins wiki page and close out this issue.

hyPiRion:
Thanks for looking into this! Based on the points brought up here, I'll guess that this is something we will never end up supporting officially, because we'd like to be compatible with the general JAR/Maven scheme used.

I have added lein-release-4digit-version to the deployment section on the Plugins wiki page though, which should be sufficient for those in need: https://github.com/technomancy/leiningen/wiki/Plugins#deployment

cprice404:
[...]

PR #2050, opened 2015-12-24
(#2043) `release` task: support SNAPSHOTs on qualifiers

cprice404:
Prior to this commit, the `lein release` task was not
really usable for releases that had qualifiers such
as `-alpha1`, `-beta2`, or `-RC1`.

The reason for this is that the code for parsing the
semver could not handle a qualifier like `-alpha1`
AND a `-SNAPSHOT` qualifier at the same time.

The default `:release-tasks` include a `bump-version`
at the `:release` level, and then the release, and
then a `bump-version` at the `:patch` level.  If
you had a version that started out as `1.0.0-alpha1` *or*
`1.0.0-alpha1-SNAPSHOT`, then the `:release` bump
would always set the version to `1.0.0`.  Then the
actual release would occur, and then the `:patch` bump
would take you to `1.0.1-SNAPSHOT`.

With this commit, `qualifier` and `snapshot` are
separated into two separate fields in the version map.
This allows us to modify the `bump-version` behavior
so that it works as one would expect with a qualifier
like `alpha`:

```clj
(is (= (bump-version "1.0.0-alpha1-SNAPSHOT" :release)
       "1.0.0-alpha1"))
(is (= (bump-version "1.0.0-alpha1")
       "1.0.0-alpha2-SNAPSHOT"))
```

This allows the default `:release-tasks` setup to be
used with qualifiers like alpha/beta.

The commit also adds a new `level` that can be passed
to the `change` task: `:qualifier`.  If this level
is passed, then the qualifier will be incremented
instead of the major/minor/patch.  It is a superset
of the existing alpha/beta/RC behavior but works with
arbitrary qualifier strings.

### DIFF
--- a/src/leiningen/release.clj
+++ b/src/leiningen/release.clj
@@ -4,20 +4,22 @@
             [leiningen.core.main :as main]
             [leiningen.core.project :as project]))
 
-(def ^:dynamic *level* :patch)
+(def ^:dynamic *level* nil)
 
 (defn string->semantic-version [version-string]
   "Create map representing the given version string. Returns nil if the
   string does not follow guidelines setforth by Semantic Versioning 2.0.0,
   http://semver.org/"
-  ;; <MajorVersion>.<MinorVersion>.<PatchVersion>[-<BuildNumber | Qualifier >]
-  (let [version-map (->> (re-matches #"(\d+)\.(\d+)\.(\d+).*" version-string)
-                         (drop 1)
-                         (map #(Integer/parseInt %))
-                         (zipmap [:major :minor :patch]))
-        qualifier (last (re-matches #".*-(.+)?" version-string))]
-    (if-not (empty? version-map)
-      (merge version-map {:qualifier qualifier}))))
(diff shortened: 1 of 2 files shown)


arichiardi:
:+1: I will test it as well but you've pretty much already done it ;)

hyPiRion:
Thanks!

This seems to be good, but I'm in the middle of vacation right now so I can't look at it thoroughly. I'll probably be able to look at it during the weekend.

cprice404:
ping!  just got back from vacation myself, just checking in to make sure this didn't slip through the cracks.  hoping to make sure that whatever I do for #2049 is compatible with whatever this one ends up looking like :)

hyPiRion:
Yeah, thanks for the ping :) I got slightly more "busy" than anticipated during that weekend, as one always seem to do, so sorry about that.

You've done really good work with this patch, thank you for that. Reads well (both the description and the code), works as intended and I can't see anything that warrants any discussion. =)

cprice404:
awesome, thanks!

arichiardi:
Hey guys, have you seen [this](https://github.com/technomancy/leiningen/issues/2060), is it just me?

PR #2063, opened 2016-01-11
Gpg piping fix

### LINKED ISSUE #2062: GPG credentials decrypting results in a NullPointerException
I have noticed that I cannot execute a `lein deploy clojars` with a standard `credentials.clj.gpg` anymore on master. I have investigated and tracked the error to https://github.com/technomancy/leiningen/commit/fb876c7d49201e4e1253e271abe4b4086a20439d.

Basically what happens now is that in case of `gpg` program succeeding, the decrypted `edn` is printed in the terminal (at least here) and [`out`](https://github.com/technomancy/leiningen/commit/fb876c7d49201e4e1253e271abe4b4086a20439d#diff-ad66867b6d37245a7ff05811484b4624R143) results in a nil with consequent `NullPointerException`.
I don't think the employed `waitFor` returns more than the exit code and therefore [the following destructuring is not accurate](https://github.com/technomancy/leiningen/commit/fb876c7d49201e4e1253e271abe4b4086a20439d#diff-ad66867b6d37245a7ff05811484b4624R136).

I was able to solve it by modifyging `gpg` as follows, but of course I might be missing the reason why `Pipe` was introduced in the first place.

```
(defn gpg
  "Shells out to (gpg-program) with the given arguments"
  [& args]
  (try
    (println "GPG program" (gpg-program))
    (let [proc-env (as-env-strings (get-english-env))
          proc-args (into-array String (concat [(gpg-program)] args))
          proc (.exec (Runtime/getRuntime) proc-args proc-env)]
      (.addShutdownHook (Runtime/getRuntime)
                        (Thread. (fn [] (.destroy proc))))
      (with-open [out (.getInputStream proc)
                  err (.getErrorStream proc)]
        (let [#_pump-out #_(doto (Pipe. out System/out) .start) <-- commented out
              pump-err (doto (Pipe. err System/err) .start)]
          #_(.join pump-out)                                    <-- commented out
          (.join pump-err)
          (let [exit-code (.waitFor proc)]
            {:exit exit-code :out (slurp (InputStreamReader. out))})))) <- out slurped and sent as `:out`
    (catch java.io.IOException e
      {:exit 1 :err (.getMessage e)})))
```

I am willing to complete this task with a little more background and for this reason I am pinging @hyPiRion :smiley:

arichiardi:
This fixes #2062 as discussed in the issue.

Also, I `*read-eval*` is now bound to false when reading (untrusted) `credendials.clj.gpg`.

### DIFF
--- a/leiningen-core/src/leiningen/core/user.clj
+++ b/leiningen-core/src/leiningen/core/user.clj
@@ -139,7 +139,8 @@
            (println "Could not decrypt credentials from" (str file))
            (println err)
            (println "See `lein help gpg` for how to install gpg."))
-         (read-string out)))))
+         (binding [*read-eval* false]
+           (read-string out))))))
 
 (def credentials (memoize credentials-fn))
 
--- a/leiningen-core/src/leiningen/core/user.clj
+++ b/leiningen-core/src/leiningen/core/user.clj
@@ -5,7 +5,8 @@
             [clojure.java.shell :as shell]
             [leiningen.core.utils :as utils])
   (:import (com.hypirion.io Pipe)
-           (java.util.regex Pattern)))
+           (java.util.regex Pattern)
+           (java.io InputStreamReader)))
 
 (defn getprop
   "Wrap System/getProperty for testing purposes."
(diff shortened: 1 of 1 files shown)


hyPiRion:
Looks good, thanks :smiley:

PR #2070, opened 2016-01-21
Add docs for providing :jvm-opts

danielcompton:
I wasn't quite sure where to put this. TUTORIAL seemed reasonable, though I'd be open to other places to, or maybe a new docs section explaining in more detail Leiningen options?


hyPiRion:
Thanks! I think this is fine in TUTORIALS.md as of right now.

It may be interesting to pull out project options into its own document, but that would require a considerable amount of effort and I'd rather not do that just yet.

PR #2071, opened 2016-01-23
Remove clojars section from GPG instructions

tobias:
The clojars section is out of date - clojars no longer has promotion to a releases repo, so no longer needs to store the user's public key. See https://github.com/clojars/clojars-web/issues/415.


hyPiRion:
Oh yeah, I read that in the announcements but forgot that we had documentation on that; Good catch.

PR #2075, opened 2016-02-01
Upgrade clojure-complete from 0.2.3 to 0.2.4

danielcompton:
0.2.4 contains type hints to eliminate reflection warnings. The diff between 0.2.3 and 0.2.4 is https://github.com/ninjudd/clojure-complete/compare/0.2.3...0.2.4. It doesn't look like there's anything too objectionable in there.


hyPiRion:
Ah, thanks. Yeah, this should be fine to bump without any worries.

PR #2032, opened 2015-11-24
Upgrade clj-http to 2.0.0

danielcompton:
This makes a lot of it's dependencies optional, including tools.reader.

Issue #1693, opened 2014-09-15
potemkin throws ExceptionInitializerError on deploy


technomancy:
https://gist.github.com/technomancy/ca0c5422b63c1f89dd3d

Not sure what's going on here, but ideally we could remove the potemkin dependency.

hyPiRion:
I've experienced this lately with `bin/lein test`, but with complaints on not finding `clojure.lang.PersistentUnrolledVector` (found in clj-tuple, which potemkin depends on). We may have some transitive version range dependency that dies because clj-http gets to decide on an older version of potemkin. I decided to pin it, hopefully this issue will be resolved as well.

hyPiRion:
I am unable to reproduce this now, so I'm guessing most of this would be related to bumping clj-http (#1959) and/or figure out what we need to do with potemkin.

PR #2076, opened 2016-02-01
Remove potemkin dependency

danielcompton:
It doesn't seem like Potemkin is being used anywhere in Leiningen, so it could be removed?

There would probably need to be a warning in the upgrade process, as some plugins may have transitive dependencies on Potemkin.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -25,7 +25,6 @@
                  [slingshot "0.10.3"]
                  [cheshire "5.5.0"]
                  [clj-http "2.0.0"]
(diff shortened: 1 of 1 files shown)


hyPiRion:
This is effectively a better fix for #1693 which now seems to be sound due to #2032: We use clj-http 2.0.0 which uses the right version of potemkin now. Previously we had to pin potemkin partially because of clj-tuple issues.

Thanks for bringing this up again so that we could finally fix that one :)

PR #2078, opened 2016-02-03
Add troubleshooting steps for gpg: decryption failed

danielcompton:
Running these steps helped me realise I had encrypted my
~/.lein/credentials.clj.gpg file with the wrong key.


hyPiRion:
Thanks! Documentation and pointers about errors hard to debug is very welcome.

danielcompton:
Yeah, that error message seems to come up for several different cases. I went down a wild goose chase with gpg-agent for a while.

PR #2088, opened 2016-02-13
Git push will push only branch related tags

### LINKED ISSUE #1846: git implementation of leiningen.vcs/push indiscriminately pushes tags
https://github.com/technomancy/leiningen/blob/master/src/leiningen/vcs.clj#L56

This is particularly problematic in CI contexts where a CI system may create CI-specific tags for use within CI "pipelines".

kleewho:
fix #1846


hyPiRion:
Thanks for the patch; looks good!

PR #2090, opened 2016-02-15
Honor whitelist settings when javac called via jar - fixes #2089

### LINKED ISSUE #2089: settings such as `local-repo` and `mirrors` from user profile not honored when `javac` is called by `jar`
We noticed an unusual thing in our CI jobs recently;  all of our CI jobs have a `~/.lein/profiles.clj`, which contains a `user` profile, which provides values for `:local-repo` and `:mirrors`.  We use `:local-repo` to keep the artifacts localized to the specific CI job (rather than system-wide), and we use `:mirrors` to cause all of the jobs to use our local nexus mirror rather than hammering on clojars and maven central every time they run.

However, we noticed from watching the logs of these jobs that they were often downloading all of the dependency artifacts twice(!); once from our nexus mirror (and storing them in `:local-repo`), and then a second time (from clojars / maven central, bypassing our nexus mirror, and storing the artifacts in ~/.m2).

After some poking I determined that this was happening on projects that contained Java sources.  It happens when the `jar` task is setting up the `eval-in-subprocess` call for `javac`.  `jar` calls `javac` via some "prep tasks" stuff, and passes a project map that has the `user` profile set up to be excluded.  Then the `javac` code does some manipulation to the project related to profiles, which involves "un-merging" the excluded profiles (including user):

https://github.com/technomancy/leiningen/blob/2ca63f0084091d7799482ebe97cb60b8f17d76c9/src/leiningen/javac.clj#L120

At that point, settings like `:local-repo` and `:mirrors` that were defined in the `user` profile are no longer in the project map.  Then, in preparing a shell command to run javac in, we hit a line of code that ends up doing classpath resolution:

https://github.com/technomancy/leiningen/blob/2ca63f0084091d7799482ebe97cb60b8f17d76c9/leiningen-core/src/leiningen/core/eval.clj#L252

This causes all of the dependencies to be resolved again without respecting the `:local-repo` and `:mirrors` settings from the `user` profile.

cprice404:
This commit refactors a few things from the `jar` and
`javac` tasks in order to ensure that the "whitelisted"
settings from the user profile (`:local-repo`, `:mirrors`,
etc.) are honored when the `jar` task launches `javac`.

It also adds a test to validate the new behavior.

### DIFF
--- a/test/leiningen/test/jar.clj
+++ b/test/leiningen/test/jar.clj
@@ -3,10 +3,11 @@
             [leiningen.core.main :as main]
             [leiningen.core.project :as project]
             [leiningen.core.utils :refer [platform-nullsink]]
-            [leiningen.test.helper :as helper])
+            [leiningen.test.helper :as helper]
+            [robert.hooke :as hooke]
+            [leiningen.javac :as javac])
   (:use [clojure.test]
-        [leiningen.jar])
-  (:import (java.util.jar JarFile)))
+        [leiningen.jar]))
 
 (def long-line
   (apply str (repeat 10000 "a")))
@@ -134,3 +135,23 @@
                                  :bytes ""}]))]
       (is (not (.contains out-str
                           "Warning: The Main-Class specified does not exist"))))))
+
+(deftest javac-launched-with-whitelisted-settings
+  (let [user-profile {:local-repo "foo/bar"
+                      :mirrors {"central" {:name "central"
+                                           :url "http://uk.maven.org/maven2"}}}
+        orig-project (-> (helper/read-test-project-with-user-profiles
+                          "java-main"
+                          {:user user-profile}))
+        javac-project (atom {})
+        javac-project-hook (fn [f proj profile]
+                             (let [new-project (f proj profile)]
+                               (reset! javac-project new-project)
+                               new-project))]
+    (hooke/with-scope
+     (hooke/add-hook #'javac/javac-project-for-subprocess javac-project-hook)
+     (jar orig-project)
+     (is (= (:local-repo orig-project)
+            (:local-repo @javac-project)))
+     (is (= (:mirrors orig-project)
+            (:mirrors @javac-project))))))
--- a/leiningen-core/src/leiningen/core/project.clj
+++ b/leiningen-core/src/leiningen/core/project.clj
@@ -837,6 +837,16 @@
                           :excluded-profiles exclude-profiles
                           :profile-inherited-meta include-profiles-meta}))))
 
+(def whitelist-keys
+  "Project keys which don't affect the production of the jar (sans its name)
+  should be propagated to the compilation phase and not stripped out."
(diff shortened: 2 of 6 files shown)


cprice404:
This fixes #2089 for me.

hyPiRion:
Thanks for the PR =)

Seems like some existing tests with a cold cache broke. I'll have a look at it when I get the time, but I'd obviously be happy to get help with that if someone has time to look at it.

cprice404:
hmmm, crud.  the tests passed for me locally.  Will try to repro the failure whenever I get a chance.

cprice404:
It actually looks like it's my new test that is failing, and I am able to repro it locally now (on a different box from where I did the original work.  computers are awesome).  Will see if I can put up a fix.

cprice404:
It looks like my test was using a maven2 mirror that no longer actually maintains any content.  I switched the test to use another mirror that actually has the clojure jar on it, and squashed that into my previous commit.  Tests pass for me locally now.

test/leiningen/test/jar.clj
```diff
@@ -134,3 +135,23 @@
                                  :bytes ""}]))]
       (is (not (.contains out-str
                           "Warning: The Main-Class specified does not exist"))))))
+
+(deftest javac-launched-with-whitelisted-settings
+  (let [user-profile {:local-repo "foo/bar"
+                      :mirrors {"central" {:name "central"
+                                           :url "http://uk.maven.org/maven2"}}}
```
timvisher: Is it just me or does relying on a maven repository on the internet seem like a bad idea in a unit test? Unless I'm misunderstanding the comments on the pr in general and this is just a value that could be arbitrary.

cprice404: I share the concern, but there are many tests in this repo that already rely implicitly on clojars and the "main" maven central repo.  You are correct that this is adding a dependency on one more internet site.  I'm open to other ideas on how to avoid that additional constraint in the test, but since a core part of leiningen is to resolve dependencies from the internets, and the bug that I opened was directly related to its ability to do that, the pros of having test coverage seemed to outweigh the cons of relying on one more mirror.

But, again, totally happy to revisit that if anyone has any better ideas!

timvisher: Forgive me for butting in on something that I don't quite understand. :)

What about a `with-redefs` form that overrides the internal function that would fetch the repo from the internet and simply hard codes the return value?

hyPiRion: Yeah, one should consider the leiningen tests as both unit tests and integration tests. It's really hard to test mirrors without actually using mirrors.

timvisher: kk. carry on. :)

hyPiRion:
Thanks! =) This looks good – only thing that I guess may cause breakage here is moving `whitelist-keys` out of the jar namespace. I'll add in a def there with reference to the original thing to ensure backwards compatibility.

PR #2092, opened 2016-02-16
Fix file handle leak in pom.properties FAQ

timvisher:
There was a file handle leak in the code as it was. It would never close the open reader against the jar resource. This will eventually cause JVM to be unable to open new files.

Functional Tests
----------------------

I'm using the code as I wrote it in my own project and there is now no file handle leak.

### DIFF
--- a/doc/FAQ.md
+++ b/doc/FAQ.md
@@ -144,8 +144,9 @@
   your project):
 
 ```clj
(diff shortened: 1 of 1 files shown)


hyPiRion:
Oh, nice catch – thanks for the patch =)

PR #2094, opened 2016-02-18
Add debug-mode warning when files are dropped from uberjar due to name clash

pkan:
When joining multiple jars into an uberjar, files in different jars may share the same path. Currently, Leiningen keeps the first, and silently discards the rest of such clashing entries.  Done in reverse dependency order, main project entries take precedence over ones from dependencies, which is good.

This behavior can be overridden with:
- `:uberjar-exclusions` feature, specifying which paths not to include
- `:uberjar-merge-with` feature, allowing for a user-defined merging of the clashing entries

There are situations when the user may be unaware that files are being discarded, when they instead _have to_ be merged.  A case in point is when multiple [CLJSJS](https://http://cljsjs.github.io/) libraries are included as project dependencies and left on the classpath for use by the ClojureScript compiler at application runtime.  The `deps.cljs` files from each lib ought to be merged, similarly to `leiningen.uberjar/clj-map-merger` but using `(partial merge-with into)` to join the data.  The same sort of problem may occur with `data_readers.clj`.  This creates a situation where code that works in development, breaks when deployed as uberjar.  So it would be very useful to issue a warning about which files are being discarded.

Because filenames are not exposed via `:uberjar-merge-with` feature, the warning is best implemented in Leiningen itself.

Since this has the potential to clutter the output with lots of new lines, the warning should be limited to when the debug environment variable is set.  To make the feature more useful, the warning should be suppressed for folders, `project.clj`, manifests, and other frequent files that are definitely not runnable code, and whose omission is unlikely to break things.  This is what this pull request proposes to do.

The message I used is: "Dropping " filename, indented two spaces for visual grouping under the jar name.  I did not use the popular term _duplicate_, as that can leave a false impression that the contents of the clashing files have been compared and found identical.

The feature could be made less verbose and still useful, by limiting the warnings to `.clj` extension and its dialects.  However, I feel that `.class` files, which tend to trigger the bulk of the warnings, are important enough to know about, since libraries distributed with aot-compiled classes have the potential to cause version conflicts unsolvable with Leiningen.

### DIFF
--- a/src/leiningen/uberjar.clj
+++ b/src/leiningen/uberjar.clj
@@ -102,6 +102,11 @@
   (or (->> mergers (filter #(merger-match? % filename)) first second)
       default-merger))
 
+(defn- warn-on-drop [filename]
+  (let [non-code #".*/|project\.clj|META-INF/(MANIFEST\.MF|(NOTICE|LICENSE)(.*\.txt)?|DEPENDENCIES)"]
+    (if-not (re-matches non-code filename)
+      (main/debug "  Dropping" filename))))
+
 ;; TODO: unify with copy-to-jar functionality in jar.clj (for 3.0?)
 (defn- copy-entries
   "Read entries of ZipFile `in` and apply the filename-determined entry-merging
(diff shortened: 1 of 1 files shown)


hyPiRion:
Hey there,

Thanks for the pull request and the well-written report!

I think this is a nice addition and I believe it would work just fine as it is is written now. If it seems to verbose in practice, we can consider limiting it to different types as you wrote.

PR #2100, opened 2016-02-23
Allow user repositories with uberjar Fixes #2098

### LINKED ISSUE #2098: uberjar task does not include :repositories from ~/.lein/profiles.clj
I have encountered what I think is a strange behaviour regarding uberjars and dependency resolution of artifacts hosted on private repositories.

Firstly (unsurprisingly) `:repositories` mentioned in your `project.clj` work as expected i.e. running `lein uberjar` and `lein test` etc... will correctly fetch the dependencies from the specified `private-repo`.  (I mention this simply to be clear that there aren't problems elsewhere with my config)

````clojure
(defproject example "0.0.1-SNAPSHOT"

  ;; ...
  ;; This works
  :repositories [["snapshots" {:url "s3p://private-repo/snapshots/"
                               :sign-releases false
                               :username :env
                               :passphrase :env
                               :releases false}]
                 ["releases" {:url "s3p://private-repo/releases/"
                              :sign-releases true
                              :username :env
                              :passphrase :env
                              :snapshots false}]]

  :plugins [[s3-wagon-private "1.2.0" :exclusions [commons-logging commons-codec]]])
````

If you then move the `:repositories` definition into your `~/.lein/profiles.clj`, under the `:user` profile then you will find most leiningen tasks will work as expected; however `lein uberjar` will fail to resolve the dependencies it requires.  e.g.

With the above `project.clj` sans the `:repositories` and the following `~lein/profiles.clj`:

````clojure
{:user
  {
   :repositories [["snapshots" {:url "s3p://swirrl-jars/snapshots/"
                                :sign-releases false
                                :username :env
                                :passphrase :env
                                :releases false}]
                  ["releases" {:url "s3p://swirrl-jars/releases/"
                               :sign-releases true
                               :username :env
                               :passphrase :env
                               :snapshots false}]]}
 :plugins [[s3-wagon-private "1.2.0" :exclusions [commons-logging commons-codec]]]
}
````

You receive errors like this:

````
(:repositories detected in user-level profiles! [:user]
See https://github.com/technomancy/leiningen/wiki/Repeatability)
Could not find artifact foo:bar:jar:0.3.5 in central (https://repo1.maven.org/maven2/)
Could not find artifact foo:bar:jar:0.3.5:jar:0.3.5 in clojars (https://clojars.org/repo/)
This could be due to a typo in :dependencies or network issues.
If you are behind a proxy, try setting the 'http_proxy' environment variable.
Uberjar aborting because jar failed: Could not resolve dependencies
````

However tasks like `lein test` `lein repl` etc... will work as expected.

Defining a custom `:uberjar` profile in `profiles.clj` does not appear to work either.  It appears that uberjar will not merge the `:repositories` key from `profiles.clj`.

After a long time playing with this, the only setup I have found that works with uberjar is either coding the repositories directly into the top level `project.clj` or the following:

`project.clj`:

````clojure
(defproject example "0.0.1-SNAPSHOT"

  ;; ...

  :profiles {:uberjar [:private-repo] }

  )
````

`profiles.clj`:

````clojure
{

 :private-repos {
                 :repositories [["snapshots" {:url "s3p://private-jars/snapshots/"
                                              :sign-releases false
                                              :username :env
                                              :passphrase :env
                                              :releases false}]
                                ["releases" {:url "s3p://private-jars/releases/"
                                             :sign-releases true
                                             :username :env
                                             :passphrase :env
                                             :snapshots false}]]
                 :plugins [[s3-wagon-private "1.2.0" :exclusions [commons-logging commons-codec]]]}

 :user [:private-repos
        {
         ;;...
         }]
 }
````

I don't know what causes this, but it looks like it might be related to either the `:repositories` key not being in the [whitelisted-keys](https://github.com/technomancy/leiningen/blob/master/src/leiningen/jar.clj#L223) or perhaps related to the [uberjar not resolving from profile vector](https://github.com/technomancy/leiningen/issues/1833).  Also I noticed [this comment](https://github.com/technomancy/leiningen/blob/master/leiningen-core/src/leiningen/core/project.clj#L192) about moving `:repositories` into the defaults project in leiningen 3.0, again I don't know if this issue is actually related to any of these things, but thought it might help to triage some potentially related findings here.

Regardless, I can't think of any reason why the uberjar task should not resolve `:repositories` from user defined profiles, it feels like a bug though I could be wrong.

Thanks again for all your hard work on leiningen!

RickMoynihan:
Allow user defined private repositories when building jars, uberjars and running deploy tasks.

For details see issue at:

https://github.com/technomancy/leiningen/issues/2098

### DIFF
--- a/leiningen-core/src/leiningen/core/project.clj
+++ b/leiningen-core/src/leiningen/core/project.clj
@@ -840,7 +840,7 @@
 (def whitelist-keys
   "Project keys which don't affect the production of the jar (sans its name)
   should be propagated to the compilation phase and not stripped out."
-  [:offline? :local-repo :certificates :warn-on-reflection :mirrors :uberjar-name :jar-name])
+  [:certificates :jar-name :local-repo :mirrors :offline? :repositories :uberjar-name :warn-on-reflection])
 
 (defn retain-whitelisted-keys
   "Retains the whitelisted keys from the original map in the new one."


hyPiRion:
This looks good; thanks for the patch!

Issue #2073, opened 2016-01-24
lein search fails when downloading clojars.org index


wbthomason:
I've recently made fresh installs of leiningen on machines running OSX and Arch Linux, using their respective package managers (homebrew and the AUR). When I run `lein search ring`, leiningen updates its package indices, as expected. However, the program appears to deadlock or hang on "Downloading https://clojars.org/repo/.index/nexus-maven-repository-index.properties". 

By looking in .lein, I've confirmed that there is a folder made for the clojars.org index, and it appears to have all the necessary files other than `timestamp`, based on a loose comparison to the directory created for the maven.org repo. If I copy `timestamp` over to the clojars folder, `lein search` no longer hangs, but it also doesn't appear to search the clojars.org repo (as evinced by the lack of any reference to ring-clojure/ring in the results for `lein search ring`. Further, by looking at network traffic while the initial `lein search` is running, it seems that it only downloads 4KB in total, which seems to indicate that it isn't downloading anything other than the .properties file. 

This seems like the kind of error which would either be highly reported or due to problems with system configuration on my end. As it doesn't appear to have been reported (that I could find), it seems that it may be a problem with my system. However, given that I've observed it across 3 separate machines running 2 different OSes, and @aml3 has observed the same on his Arch machine, it seems reasonable that it may be a larger issue. I've also tried letting this run overnight, in case it was just an issue of speed (as in #534), but with no change. All other functionality of leiningen that I've tested (including `lein deps`) has seemed to work correctly.

This is the output of the command on OSX:
```
lein search ring
WARNING: update already refers to: #'clojure.core/update in namespace: clj-http.client, being replaced by: #'clj-http.client/update
Updating the search index. This may take a few minutes...
Downloading https://repo1.maven.org/maven2/.index/nexus-maven-repository-index.properties
100.0% complete
Downloading https://repo1.maven.org/maven2/.index/nexus-maven-repository-index.gz
100.0% complete
Updating the search index. This may take a few minutes...
Downloading https://clojars.org/repo/.index/nexus-maven-repository-index.properties

```
When I use Ctrl+\ to get a stack trace, it has this output:
```
^\2016-01-24 13:01:09
Full thread dump Java HotSpot(TM) 64-Bit Server VM (25.40-b25 mixed mode):

"Index-BottleWarmer-clojars" #15 daemon prio=5 os_prio=31 tid=0x00007fb923800000 nid=0xae03 waiting on condition [0x0000700001558000]
[...]

PR #2101, opened 2016-02-26
Fix for issue #2073

pauldorman:
This seems to work for me.


hyPiRion:
This also seems to work for me, so thanks!

PR #2104, opened 2016-03-04
Quote CLASSPATH while calling Java Fixes #2086

### LINKED ISSUE #2086: lein.bat doesn't correctly handle whitespace in the standalone.jar file path
Running the lein batch file results in the message
`Error: Could not find or load main class Lastname\.lein\self-installs\leiningen-2.6.1-standalone.jar`
Since the full path is "C:\Users\Firstname Lastname\.lein\self-installs\leiningen-2.6.1-standalone.jar" it looks like the space in the path is throwing it off.
Putting quotes around `%CLASSPATH%` on line 380 seems to solve the issue. However, these quotes were taken out [in this commit](https://github.com/technomancy/leiningen/commit/927b9bf26c63be553a52f27281eb5185f382e78a) to fix another issue.

mastersign:
to support spaces in LEIN_JAR, the later concatenated CLASSPATH must be quoted

### DIFF
--- a/bin/lein.bat
+++ b/bin/lein.bat
@@ -380,7 +380,7 @@ set RC=0
 "%LEIN_JAVA_CMD%" -client %LEIN_JVM_OPTS% ^
  -Dclojure.compile.path="%DIR_CONTAINING%/target/classes" ^
  -Dleiningen.original.pwd="%ORIGINAL_PWD%" ^
- -cp %CLASSPATH% clojure.main -m leiningen.core.main %*
+ -cp "%CLASSPATH%" clojure.main -m leiningen.core.main %*
 SET RC=%ERRORLEVEL%
 if not %RC% == 0 goto EXITRC
 
--- a/bin/lein.bat
+++ b/bin/lein.bat
@@ -95,9 +95,9 @@ rem the paths inside the bootstrap file do not already contain double quotes but
     if "x!LEIN_LIBS!" == "x" goto NO_DEPENDENCIES
 
 
-	rem CLASSPATH must contain each path element double quoted and separated by semicolons ie. "c:\some folder";"c:\some other folder";"c:\semicolons;;;;in;name;work okay"
+	rem semicolons in pathes are not supported, spaces are supported by quoting CLASSPATH as a whole
 	rem (no end semicolon required)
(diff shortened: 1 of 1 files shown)


winks:
Autolinking in the topic doesn't work. This is a fix for #2086

mastersign:
Sorry, was not thorough enough. Feeling stupid. Feel free to reject to pull request.

hyPiRion:
@mastersign: For what it's worth, this _seems_ sensible to me and it is is what we initially did. And we don't really have a good idea for why this makes people unable to build Leiningen from Windows.

My guess is that this patch is fine and works as intended, but that this causes some other bug to show up because of that. So please, don't feel stupid: I'm thankful that you took time to look into this.

mastersign:
I think I found a possible problem in the comment in [line 98](https://github.com/mastersign/leiningen/blob/mastersign-patch-path-spaces/bin/lein.bat#L98). It states that each path in CLASSPATH must be quoted individually. This of course collides with quotes around CLASSPATH as a whole. But I think it would be sufficient, if the paths in CLASSPATH would not be quoted but instead CLASSPATH as a whole would be quoted.

As far as I can see, there are two solutions:

1. Remove the individual quotes in [line 100](https://github.com/mastersign/leiningen/blob/mastersign-patch-path-spaces/bin/lein.bat#L100) and put quotes around CLASSPATH in [line 383](https://github.com/mastersign/leiningen/blob/mastersign-patch-path-spaces/bin/lein.bat#L383)
2. Or put quotes around LEIN_JAR in [line 118](https://github.com/mastersign/leiningen/blob/mastersign-patch-path-spaces/bin/lein.bat#L118)

mastersign:
I would prefer the first solution, because the concatenation of CLASSPATH in different places in the script would get easier.

In the second solution, in *every* case where a path or path list is added to CLASSPATH, *each* path must be quoted individually. This is less maintainable.

mastersign:
The first solution drops support for semicolons in paths.

hyPiRion:
Speaking as a person with an apostrophe in their name: I think semicolon in paths is such a rare thing that you sort-of would _expect_ this to crash. I'm not saying that it's good, but I think maintainability of this .bat-script is more important right now.

PR #2106, opened 2016-03-05
Fixed character escaping issue

solatis:
This fixes the issue where, if there would be a space in the classpath, the -cp variable passed to `java` would be incorrect.

This was an issue I hit upon, since there is a space in my windows username, so that I would get an issue such as:

```
Error: Could not find or load main class Mergen\.lein\self-installs\leiningen-2.6.1-standalone.jar
```

where it is obvious that the first part of the path was cut off due to missing quotes.


hyPiRion:
Thanks for this! I'll get to verify this when I've got my Windows VM up and running, along with other Windows-related issues.

PR #2112, opened 2016-03-15
javac supports relative source paths as arguments

djsweet:
As specified in MIXED_PROJECTS.md, this is done with the intent to
support multi-stage source interleaving between Clojure and Java.

Also changed in this patchset:
- @-arguments passed to javac are now expanded
- compilation of Java sources can be avoided even when not stored
  in a directory format mirroring the package format

### DIFF
--- a/src/leiningen/javac.clj
+++ b/src/leiningen/javac.clj
@@ -6,19 +6,151 @@
             [leiningen.core.project :as project]
             [clojure.java.io :as io]
             [clojure.string :as string])
-  (:import java.io.File
+  (:import [java.io BufferedReader File FileReader]
+           java.util.regex.Matcher
            javax.tools.ToolProvider))
+
+(defn- rebuild-line
+  "\"Rebuilds\" a line of Java, so that the comments aren't present."
+  [start-line from-comment-block?]
+  (loop [rest-line                 start-line
+         in-comment?               from-comment-block?
+         ^StringBuilder return-val (StringBuilder.)]
+    (cond (empty? rest-line)
+          [(.toString return-val) in-comment?]
+          in-comment?
+          (if (and (>= (count rest-line) 2)
+                   (= (first rest-line) \*)
+                   (= (first (rest rest-line)) \/))
+            (recur (rest (rest rest-line))
+                   false
+                   return-val)
+            (recur (rest rest-line)
+                   true
+                   return-val))
+          (and (>= (count rest-line) 2)
+               (= (first rest-line) \/)
+               (= (first (rest rest-line)) \/))
+          [(.toString return-val) false]
+          (and (>= (count rest-line) 2)
+               (= (first rest-line) \/)
+               (= (first (rest rest-line)) \*))
+          (recur (rest (rest rest-line)) ;; "eat" the asterisk so that
+                 ;; /*/ doesn't happen
+                 true
+                 return-val)
+          :else
+          (recur (rest rest-line)
+                 false
+                 (.append return-val (first rest-line))))))
+
+(defn uncommented-lines-from-reader
+  "Returns a lazy sequence of lines from a Java character source
+  with the comments stripped out. Closes the reader when there is
+  no more input to read."
+  ([reader] (uncommented-lines-from-reader reader false))
+  ([^BufferedReader reader in-comment-block?]
+   (let [next-line (.readLine reader)]
+     (cond (nil? next-line)
+           (do (.close reader)
+               nil)
+           in-comment-block?
+           (if-not (.contains next-line "*/")
+             (lazy-seq (cons "" (uncommented-lines-from-reader reader true)))
+             (let [[rebuilt-line started-comment-block?]
+                   (rebuild-line next-line true)]
+               (lazy-seq (cons rebuilt-line
+                               (uncommented-lines-from-reader
+                                reader
+                                started-comment-block?)))))
+           :else
+           (if (and (not (.contains next-line "/*"))
+                    (not (.contains next-line "//")))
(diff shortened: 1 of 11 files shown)


hyPiRion:
Thanks for this! It's a big one, but it's on my prioritised list of things to take a look at.

cprice404:
I'm planning on seeing if I can grok this today, since I know that @hyPiRion would like to see it included in 2.7.0.  However it looks like it's going to need a rebase before it can be merged.

test_projects/alternating-language-compilation/LICENSE
```diff
@@ -0,0 +1,214 @@
+THE ACCOMPANYING PROGRAM IS PROVIDED UNDER THE TERMS OF THIS ECLIPSE PUBLIC
+LICENSE ("AGREEMENT"). ANY USE, REPRODUCTION OR DISTRIBUTION OF THE PROGRAM
```
cprice404: Did you intend to commit this file?  Seems like it's maybe not necessary inside one of the test projects, and might cause confusion as to which is the actual license file for leiningen itself?

src/leiningen/javac.clj
```diff
@@ -6,19 +6,151 @@
...
+(defn- extract-compiled-path
+  "Generates an \"effective\" compiled path for a given .java file.
+
+  In order to do this, the passed file is opened, and scanned for the
+  first valid package declaration. The package declaration is then transformed
+  into a relative directory path.
+
+  This function assumes that javac -d has been used, which will generate
+  a directory structure in line with the package declaration. The resulting
+  return value will be relative to the argument to javac -d."
```
cprice404: I'm a bit confused by these parts, and it seems a little worrisome to introduce so much code for trying to parse java source files.  In past projects I've worked on that sort of code has had high maintenance costs.  But I'm not sure that I'm totally understanding the use case.

Based on the lein project that you have in your test, it seems like the issue is that `javac` is being called with a path like `src/java/leiningen/test/alc`, which is not the root of your java source tree.  However, it seems like you could either:

1. require these paths to be a subdirectory of a path that is registered in `java-sources` (which is true in your example), in which case you could just relativize the path passed to `javac` against the appropriate entry from `java-sources`, and you'd know the package name, or,
2. Continue to restrict calls to `javac` so that they must be given a proper java source root and must compile the whole root, and instruct users to set up multiple java source roots if they need to do the sort of phased compilation that you are doing in your test.

Option 2 actually sounds better to me; it seems like it'd be highly desirable for code that has to be compiled in a certain order to be broken out into directories that reflect that order.  But either of the two options above seems like it would allow you to achieve the phased compilation you're looking for without the need for any of the code that is parsing the java source files.

Perhaps others have different opinions, but that'd be how I'd propose to solve this problem.

src/leiningen/javac.clj
```diff
@@ -70,9 +202,21 @@
         (.write (string/join "\n" (map safe-quote files)))))
     (into-array String
                 (concat (normalize-javac-options (:javac-options project))
-                        args
+                        (filter #(or (string/starts-with? % "-")
+                                     (and (string/starts-with? % "@")
+                                          (not (string/starts-with? % "@@"))))
```
cprice404: What's the case where a `@@` would be passed?

src/leiningen/javac.clj
```diff
@@ -70,9 +202,21 @@
...
+                                args)
                         [(str "@" options-file)]))))
 
+;; seq the map so that an empty sequence returns nil
+;; we'll use this later to see if we have any source paths in the args
+(defn- source-paths-from-args [args]
+  (seq (map #(if (string/starts-with? % "@@") (.substring 1 %) %)
+            (filter #(or (and (not (string/starts-with? % "-"))
+                              (not (string/starts-with? % "@")))
+                         (string/starts-with? % "@@"))
```
cprice404: This is basically the complement of the filter you defined above, right?  Probably worth extracting a predicate function?

src/leiningen/javac.clj
```diff
@@ -136,8 +289,12 @@
...
 
 Add a :java-source-paths key to project.clj to specify where to find them.
-Options passed in on the command line as well as options from the :javac-opts
-vector in project.clj will be given to the compiler; e.g. `lein javac -verbose`.
+Options passed in on the command line which start with '-' or '@' will be
+treated as arguments to the compiler, and all arguments in the :javac-opts
+vector in project.clj will be directly given to the compiler; e.g.
+`lein javac -verbose`. Otherwise, options will be treated as Java source
+directories (relative to the project root) for the current invocation, 
+overriding the values in the :java-source-path vector.
```
cprice404: This seems like the only real API-level change here.  Just to try to think through the backward-compatibility implications of that, are there any valid use cases where someone might have been passing in a string to `javac-opts` that did *not* begin with a `-` or a `@`?

Perhaps another idea to consider here might be to allow an options map to be specified as the final argument to `javac`.  This would be easy to distinguish from any other args because it'd be a clojure map literal and could only be defined in the `project.clj`.  If the last arg is a map it could be treated specially, and you could have a key inside of it called `source-path` or similar, where you specified the particular path you wished to compile.  This would get rid of any ambiguity compared to any strings passed from the CLI without requiring any heuristics, so it seems like it'd have a stronger guarantee of backward compatibility / safety.

hyPiRion:
I think this is a bit too big to merge in for something that very few people seem to request, as well as no replies on the comments made by @cprice404. Happy to discuss this further if you still think it's worth to add in.

I'd guess if you _really_ need this, you can use lein-shell as a workaround.

Issue #2067, opened 2016-01-19
add support for 'managed' dependencies?


cprice404:
In Maven, a parent pom can specify a section called `dependencyManagement`. If this is specified, the versions of the dependencies listed in this section are used when resolving transitive dependencies. It also allows you to specify dependencies in the main sections of the pom w/o specifying a version number for them.

Exposing this functionality through leiningen would be useful for writing leiningen plugins that did things analogous to what maven's "parent poms" can do.  I'm imagining that this would probably entail adding a new key called `:managed-dependencies` and then a little work to merge that in with the regular dependencies before calling into `pomegranate`/`aether` to do the actual dependency resolution.

(Note: `pomegranate` doesn't support this yet, but I've poked at the code and it seems like it would be very easy to add.  I captured that in https://github.com/cemerick/pomegranate/issues/72 .)

If a top-level `:managed-dependencies` section didn't seem palatable, I think there are some other ways this could be implemented by just refactoring the code that determines which keys from the project get passed in to the various functions that do the dependency resolution.  Then a lein plugin could add hooks to those functions to add in additional keys that it had defined its own support for.

At this point I'm just exploring this idea in an attempt to solve some problems we're having at our company about how to manage the version numbers of super-common dependencies that end up getting used by tons of our projects.  If there was consensus that making this possible in lein was worthwhile, I'm happy to take input on what the best implementation paths might be, and put together a PR if we actually end up pushing forward on this.

hyPiRion:
It sounds like this is possible through middleware, or am I wrong here?

I think this sounds like reasonable functionality to have for these kinds of projects, but it seems like yet another concept Lein users may have to learn. So even though I think this seems useful, I'd like to avoid including it before a bigger audience would like this in. 

If it's not possible to do this through middleware, then it may be more reasonable to push it into Leiningen.

cprice404:
I definitely hear you about wanting to minimize cognitive burden for users.  In the interest of that I'd definitely be happy enough just trying to figure out a way to be able to add support for this for a plugin.

Unfortunately, I don't think it's possible through middleware given the current lein code.  "Managed" dependencies are conceptually different than regular dependencies, in that they don't actually cause a dependency to be resolved and brought into the project; they only specify a version number to be used if something *else* brings the dependency into the project.

[...]

PR #2126, opened 2016-04-10
(#2067) add support for `managed-dependencies`

cprice404:
FOR REVIEW: this PR is just up for early review.  There is still some cleanup to do, and additional tests to add.  Also, I'm planning on tackling the "parent project" work somewhat separately, but it might make sense to hold off on this one until that part is at least somewhat proven out.  I'd appreciate any feedback on this PR in its current state, though.

---------------------------------------------------------------------------------------

This commit provides initial support for `managed-dependencies`,
where dependency version numbers may be specified in a separate
section called `managed-dependencies`, and those version numbers
will be used for any deps in the main `dependencies` section
that do not explicitly specify a version number.

This is a precursor to being able to specify a "parent" project
that could be used to consolidate version numbers of common
dependencies across a large number of libraries.

### DIFF
--- a/doc/MANAGED_DEPS.md
+++ b/doc/MANAGED_DEPS.md
@@ -23,13 +23,14 @@ regular `:dependencies` section, with two exceptions:
 
 1. It does not actually introduce any dependencies to your project.  It only says,
   "hey leiningen, if you encounter one of these dependencies later, here are the
-  versions that you should use."
+  versions that you should fall back to if the version numbers aren't explicitly
+  specified."
 2. It allows the version number to be omitted from the `:dependencies` section,
   for any artifact that you've listed in your `:managed-dependencies` section.
 
 Here's an example:
 
-```clojure
+```clj
 (defproject superfun/happyslide "1.0.0-SNAPSHOT"
   :description "A Clojure project with managed dependencies"
   :min-lein-version  "2.7.0"
@@ -59,7 +60,7 @@ the [`lein-parent` plugin](https://github.com/achin/lein-parent).  This plugin
 will allow you to define a single "parent" project that is inherited by multiple
 "child" projects; e.g.:
 
-```
+```clj
 (defproject superfun/myparent "1.0.0"
    :managed-dependencies [[clj-time "0.12.0"]
                             [me.raynes/fs "1.4.6"]
--- a/doc/MANAGED_DEPS.md
+++ b/doc/MANAGED_DEPS.md
@@ -1,0 +1,104 @@
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+# Managed Dependencies With Leiningen
+
+Maven (and now Leiningen) provides a capability called "Dependency Management".
+The idea is to provide a way to specify a version number for common library
+dependencies in a single location, and re-use those version numbers from other
+discrete maven/lein projects.  This makes it easy to, e.g., update your `clj-time`
+dependency across a large number of projects without having to be mindful
+of every common dependency version across all of your libraries.
+
+When using `:pedantic? :abort` in your projects, to ensure that you are producing
+a consist and predictable build, it can be very cumbersome to play the "dependency
+version whack-a-mole" game that arises whenever an upstream library bumps a version
+of one of its dependencies.  `:managed-dependencies` can help alleviate this issue
+by allowing you to keep the dependency version numbers centralized.
+
+## `:managed-dependencies`
+
+The `:managed-dependencies` section of your `project.clj` file is just like the
+regular `:dependencies` section, with two exceptions:
+
+1. It does not actually introduce any dependencies to your project.  It only says,
+  "hey leiningen, if you encounter one of these dependencies later, here are the
+  versions that you should use."
+2. It allows the version number to be omitted from the `:dependencies` section,
+  for any artifact that you've listed in your `:managed-dependencies` section.
+
+Here's an example:
+
+```clojure
+(defproject superfun/happyslide "1.0.0-SNAPSHOT"
+  :description "A Clojure project with managed dependencies"
+  :min-lein-version  "2.7.0"
+  :managed-dependencies [[clj-time "0.12.0"]
+                         [me.raynes/fs "1.4.6"]
+                         [ring/ring-codec "1.0.1"]]
+  :dependencies [[clj-time]
+                 [me.raynes/fs]])
+```
+
+In the example above, the final, resolved project will end up using the specified
+ versions of `clj-time` and `me.raynes/fs`.  It will not have an actual dependency
+ on `ring/ring-codec` at all, since that is not mentioned in the "real" `:dependencies`
+ section.
+
+This feature is not all that useful on its own, because in the example above,
+we're specifying the `:managed-dependencies` and `:dependencies` sections right
+alongside one another, and you could just as easily include the version numbers
+directly in the `:dependencies` section.  The feature becomes more powerful
+when your build workflow includes some other way of sharing the `:managed-dependencies`
+section across multiple projects.
+
+## Lein "parent" projects
+
+One way of leveraging `:managed-dependencies` across multiple projects is to use
+the [`lein-parent` plugin](https://github.com/achin/lein-parent).  This plugin
+will allow you to define a single "parent" project that is inherited by multiple
+"child" projects; e.g.:
+
+```
+(defproject superfun/myparent "1.0.0"
+   :managed-dependencies [[clj-time "0.12.0"]
+                            [me.raynes/fs "1.4.6"]
+                            [ring/ring-codec "1.0.1"]])
+
+(defproject superfun/kid-a "1.0.0-SNAPSHOT"
+   :parent-project [:coords [superfun/myparent "1.0.0"]
+                    :inherits [:managed-dependencies]]
+   :dependencies [[clj-time]
+                  [me.raynes/fs]])
+
+(defproject superfun/kid-b "1.0.0-SNAPSHOT"
+ :parent-project [:coords [superfun/myparent "1.0.0"]
+                  :inherits [:managed-dependencies]]
+ :dependencies [[clj-time]
+                [ring/ring-codec]])
+```
+
+In this example, we've consolidated the task of managing common version dependencies
+in the parent project, and defined two child projects that will inherit those
+dependency versions from the parent without needing to specify them explicitly.
+
+This makes it easier to ensure that all of your projects are using the same versions
+of your common dependencies, which can help make sure that your uberjar builds are
+more predictable and repeatable.
+
+## Other ways to share 'managed-dependencies'
+
+Since the `defproject` form is a macro, it would be possible to write other plugins
+that generated the value for a `:managed-dependencies` section dynamically.  That
+could provide other useful ways to take advantage of the `:managed-dependencies`
+functionality without needing to explicitly populate that section in all of your
+`project.clj` files.
+
+## Future integration
+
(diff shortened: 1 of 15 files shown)


leiningen-core/src/leiningen/core/classpath.clj
```diff
@@ -486,6 +493,19 @@
           (doseq [[_ {:keys [native-prefix file]}] snap-deps]
             (extract-native-dep! native-path file native-prefix))))))
 
+(defn resolve-managed-dependencies
+  "TODO"
```
cprice404: Need to fix this

leiningen-core/src/leiningen/core/classpath.clj
```diff
@@ -486,6 +493,19 @@
...
+                                 managed-dependencies-key project rest)
+        jars (->> dependencies-tree
+                  (aether/dependency-files)
+                  (filter #(re-find #"\.(jar|zip)$" (.getName %))))]
+    (when (and (= :dependencies dependencies-key)
+               (:root project))
+      (extract-native-dependencies project jars dependencies-tree))
+    jars))
+
 (defn resolve-dependencies
```
cprice404: I left this in because I didn't want to break API, but I am wondering if it should be marked as deprecated or something since it should no longer be used internally?

leiningen-core/src/leiningen/core/classpath.clj
```diff
@@ -494,23 +514,25 @@
...
-               (:root project))
-      (extract-native-dependencies project jars dependencies-tree))
-    jars))
+  [dependencies-key project & rest]
+  (apply resolve-managed-dependencies dependencies-key nil project rest))
+
+(defn managed-dependency-hierarchy
+  "TODO"
+  [dependencies-key managed-dependencies-key project & options]
+  ;; TODO: explain private call
```
cprice404: There is a new fn in `pomegranate` `0.3.1` that I didn't realize that I would need to call in this PR, so it's marked as private in `pomegranate`.  I can open a PR against `pomegranate` to try to get it switch to public, but I'm worried that it might take a long time to get another `pomegranate` release with that change.

test_projects/managed-deps/project.clj
```diff
@@ -0,0 +1,12 @@
...
+(defproject mgmt "0.99.0-SNAPSHOT"
+  :description "A test project"
+
+  :managed-dependencies [[~(symbol "org.clojure" "clojure") ~clj-version]
+                 [rome ~(str "0." "9")]
+                 [ring/ring "1.0.0"]]
+
+  :dependencies [[org.clojure/clojure]
+                 [rome/rome nil]
+                 [ring]])
```
cprice404: I'd like to add several more things to this file, such as deps that have exclusions, classifiers, etc.

test/leiningen/test/deps.clj
```diff
@@ -128,3 +127,11 @@
...
                                                        native-project)))))]
                 (.getName f))))))
+
+(deftest ^:online test-managed-deps
+  (let [managed-deps [["rome" "0.9"] ["jdom" "1.0"]]]
+    (doseq [[n v] managed-deps]
+      (delete-file-recursively (m2-dir n v) :silently))
+    (deps managed-deps-project)
+    (doseq [[n v] managed-deps]
+      (is (.exists (m2-dir n v)) (str n " was not downloaded.")))))
```
cprice404: Would like to expand this test a bit further, along with the project file that it uses.

cprice404:
@hyPiRion if you get a chance at some point, I'd be interested to hear your thoughts on this, and input on next steps, etc.

hyPiRion:
@cprice404, thanks! And sure thing, I'll get a look whenever I have time.

cprice404:
Travis failures all seem to be issues with downloading the clojure 1.7 jar from "foo.com".  This wasn't happening to me locally, guessing some of the tests are stubbed out to try not to go out to the internet and download artifacts from maven central, and that for these tests, the fact that I changed the clojure version from 1.3 to 1.7 means that the CI system doesn't already have the jars in place?

Will try clearing my local m2 cache and seeing if I can repro locally.

cprice404:
Was able to repro travis failure locally, squashed up a commit that should fix.

cprice404:
@hyPiRion ping just to make sure this didn't slip through the cracks.  If you just haven't had time yet, no worries, I totally understand!  The life of an OSS project maintainer :)

hyPiRion:
@cprice404 No worries, this is still on my list. I've been surprisingly busy the last two-three weeks, so unfortunately I've not been able to look at pull requests.

That seems to be over though, and I did some small patches last evening. Hopefully I'll be able to go through these during this week =)

cprice404:
@hyPiRion excellent, thank you.

cprice404:
@hyPiRion any chance to take a peek at this yet?

leiningen-core/src/leiningen/core/classpath.clj
hyPiRion: Yeah, I think we can add in `^:deprecated` to this one.

I looked over at [crossclj](https://crossclj.info/fun/leiningen.core.classpath/resolve-dependencies.html) for uses, and it seems to be used outside of lein in a couple of plugins. So I think that, to avoid breakage with pluginsm, it would be beneficial to replace `nil` with `:managed-dependencies` here if `(= dependencies-key :dependencies)`.

leiningen-core/src/leiningen/core/classpath.clj
hyPiRion: It's fine to use a private one as we have full control of which pomegranate version we use. It would be beneficial to get it changed though.

hyPiRion:
Alright, got around to look at this one. The initial implementation here looks good, I think.

I the only thing that should be fixed is to avoid breakage of existing plugins that use `resolve-dependencies`, as they probably want to resolve dependencies. That would break if managed-dependencies-key is set to nil rather than `:managed-dependencies`.

cprice404:
@hyPiRion thanks very much!  I will look into the change you suggested.  Am also still soliciting another review from a co-worker, and then will start looking into the work to add support for the parent project file.

cprice404:
OK I pushed up some commits addressing the issue pointed out in review, cleaning up some TODOs, and adding better test coverage.  I think this PR is reasonably merge-able at this point, but it's not very useful w/o the additional work to add support for parent projects.  I'm planning to start working on that soon, and probably just add it as additional commits on this PR.

hyPiRion:
The implementation looks and seems good, but I haven't gone through the test cases. When I've looked over them I'll merge this one in. =)

senior:
@cprice404 I use `sample.project.clj` pretty heavily when making a `project.clj` change. Any reason not to update that so users know it's there and how to use it?

cprice404:
@senior no, that is a good suggestion.  I'll add something there and also take a poke around and see if there are any other docs that I need to mention this in.

cprice404:
@senior I added a commit with an example and some docs.

cprice404:
Also @senior did you look over my test coverage?  Anything you can think of that I'm not covering sufficiently?

doc/MANAGED_DEPS.md
```diff
@@ -0,0 +1,104 @@
...
+version whack-a-mole" game that arises whenever an upstream library bumps a version
+of one of its dependencies.  `:managed-dependencies` can help alleviate this issue
+by allowing you to keep the dependency version numbers centralized.
+
+## `:managed-dependencies`
+
+The `:managed-dependencies` section of your `project.clj` file is just like the
+regular `:dependencies` section, with two exceptions:
+
+1. It does not actually introduce any dependencies to your project.  It only says,
```
senior: I think of it as, when a version is not specified, use this one. Reading this makes me think that it's more of an override.

doc/MANAGED_DEPS.md
```diff
@@ -0,0 +1,104 @@
...
+section across multiple projects.
+
+## Lein "parent" projects
+
+One way of leveraging `:managed-dependencies` across multiple projects is to use
+the [`lein-parent` plugin](https://github.com/achin/lein-parent).  This plugin
+will allow you to define a single "parent" project that is inherited by multiple
+"child" projects; e.g.:
+
+```
```
senior: Above you're denoting the block as clojure, here it's not

test/leiningen/test/deps.clj
```diff
@@ -128,3 +129,62 @@
...
+                             (fn [dep]
+                               (some (partial coordinates-match? dep) merged-deps))
+                             managed-deps)
+        ;; deps that have classifiers
+        classified-deps (filter
+                         #(some #{:classifier} %)
+                         merged-deps)]
+    ;; make sure the sample data has some unmanaged deps, some unused managed deps,
+    ;; and some classified deps, for completeness
+    (is (not (empty? versioned-unmanaged-deps)))
```
senior: seq is preferred over (not (empty?...))

senior:
I think there might be some changes needed to the [pom generation code](https://github.com/technomancy/leiningen/blob/master/src/leiningen/pom.clj#L333). I'm not sure what the right behavior is. Probably at a minimum merging the managed dependencies in with the regular ones before generating the file

test_projects/managed-deps/project.clj
```diff
@@ -0,0 +1,18 @@
+(def clj-version "1.3.0")
+
+(defproject mgmt "0.99.0-SNAPSHOT"
+  :description "A test project"
+
+  :managed-dependencies [[~(symbol "org.clojure" "clojure") ~clj-version]
+                 [rome ~(str "0." "9")]
+                 [ring/ring "1.0.0"]
+                 [ring/ring-codec "1.0.1"]
+                 [commons-math/commons-math "1.2" :classifier "sources"]
```
senior: In the docs you talk about managed deps being particularly useful for transitive dependencies. Is that tested here?

test/leiningen/test/deps.clj
cprice404: `seq?` doesn't have the same semantics.  The `remove` above may return `'()`, which evaluates to `true` with `seq?`, which makes the assertion invalid.

cprice404: oh, `seq`, not `seq?`.  That seems significantly less readable to me but I'll change it.

cprice404:
@senior I added a commit that addresses all of your feedback *except* for the pom generation thing.  Still working on that, but if you wanted to confirm that this commit addresses the rest of it that'd be swell.

senior:
The changes look good but the managed-deps tests are failing

cprice404:
derp.  squashed a fix that will hopefully get travis passing.

cprice404:
@senior OK I pushed up another commit that adds `dependencyManagement` section to the pom.  Good catch on that.  I think I've addressed all of your feedback.

senior:
:+1:

cprice404:
@hyPiRion : @senior helped me review test cases and found some things that I've since fixed up.  I'll sit on this for a few days in case you have time / interest to take another look at it, but I'll probably merge it soon-ish since it seemed like you were mostly +1 on it.

And, TBH, if you did end up with any bandwidth for PR review, I'd love to see if you have any thoughts on #2164 since that one hasn't had as many eyes on it.  Thanks!

cprice404:
OK, went ahead and merge this since I think it's been reviewed pretty thoroughly.  @hyPiRion I'll keep an eye out for related bugs and try to make sure I address anything if it comes up.  I'm going to solicit some review from my co-workers on #2164, and once that is merged we will wait for the next lein release so that we can start getting some mileage on the `lein-parent` plugin!

cprice404:
looks like this broke the build.  I'll start working on a fix right away.

hyPiRion:
For reference: #2170 is the actual issue here – the build failure is completely unrelated to this PR.

PR #2127, opened 2016-04-13
Fix plugin profiles documentation

RasterBurn:
The documentation said to "Create a file called `src/myplugin/profiles.clj`".  This didn't work for me.  I had to create a file `resources/myplugin/profiles.clj` instead.


hyPiRion:
Nice catch, thanks :)

PR #2135, opened 2016-04-27
Add support for get-in pprinting to lein-pprint

dryewo:
Also add tests for lein-pprint.


hyPiRion:
Thanks! This seems like a useful addtion to the pprint task :)

emlyn:
This looks handy, but doesn't seem to work with the latest lein-pprint, does it need a new release?

technomancy:
Just pushed a 1.2.0 release; thanks for the heads up.

PR #2138, opened 2016-04-29
Catch and handle fixture errors during tests

alexhall:
Injected a hook in `clojure.test/test-ns` to catch and handle exceptions, which will prevent the test process from dying and allowing testing to continue on other namespaces.


hyPiRion:
Thanks! This looks good. =)

PR #2142, opened 2016-05-08
Fixed typo in tutorial

ashleyconnor:


hyPiRion:
This doesn't seem right – `require` needs its inputs quoted.

That being said, the initial namespace for a program based on the app template is `my-app.core`, so we could probably remove that part of the repl intro.

ashleyconnor:
You're right - [https://clojuredocs.org/clojure.core/require](https://clojuredocs.org/clojure.core/require).

I thought I got a different result in my repl but I must have been mistaken.

Closing this one.

PR #2146, opened 2016-05-18
Fix #2145: extra whitespace added to JVM_OPTS

### LINKED ISSUE #2145: JVM_OPTS parsing incorrectly adds trailing spaces to arguments
When the `JVM_OPTS` shell variable includes multiple spaces between arguments, the `get-jvm-opts-from-env` function incorrectly adds trailing spaces to the previous argument. For example:

```
JVM_OPTS="-Djavax.net.ssl.keyStore=/path/to/file.p12 \
          -Djavax.net.ssl.keyStoreType=pkcs12"
```

After parsing:

```
["-Djavax.net.ssl.keyStore=/path/to/file.p12         " "-Djavax.net.ssl.keyStoreType=pkcs12"]
```

This is obviously incorrect, and unfortunately very confusing (trailing whitespace is very hard to see in exceptions). The issue is caused by incorrect parameter accumulation in the `join-broken-arg` function, whitespace is added even if the given token is empty (i.e. the gap between two spaces).

sfnelson:
Instead of using String.split then recombining arguments to repair broken
quoted arguments, this parser now finds entire arguments using regex matching.
This is more robust, handling extra whitespace (and non-space whitespace)
correctly, while also correctly matching quotations in argument strings and
preserving internal spaces.

### DIFF
--- a/leiningen-core/src/leiningen/core/eval.clj
+++ b/leiningen-core/src/leiningen/core/eval.clj
@@ -110,18 +110,9 @@
 (defn- d-property [[k v]]
   (format "-D%s=%s" (as-str k) v))
 
-;; TODO: this would still screw up with something like this:
-;; export JAVA_OPTS="-Dmain.greeting=\"hello -main\" -Xmx512m"
-(defn- join-broken-arg [args x]
-  (if (= \- (first x))
-    (conj args x)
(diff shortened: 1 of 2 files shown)


hyPiRion:
Looks good, thanks! =)

Issue #2149, opened 2016-05-23
Java 9 incompatibility


tommyettinger:
Due to restructuring in the JDK's internal-use packages, `sun.misc` is very different in Java 9, and much of it is not available due to the "Project Jigsaw" modularization of the JRE actually making internal-use packages internal. This affects Lein because, at the bare minimum, `sun.misc.Launcher` is used even when running the command `lein` with no args. This class, `Launcher`, is likely being restructured because a new `sun.launcher` package is present in the root module, whichexports a small subset of `sun.misc` and other internal-use packages. This subset notably includes the `Unsafe` class, but that one may be limited to JRE builds that include the module `jdk.unsupported`, which is discouraged already and a likely candidate for removal in Java 10.

The information on this is scattered around the OpenJDK site, but this JEP is the most relevant: http://openjdk.java.net/jeps/260 , and this section on modularization should also be useful: http://openjdk.java.net/projects/jigsaw/ . I'm using the 4th preview release of the Zulu OpenJDK 9 builds, available here: http://zulu.org/zulu-9-pre-release-downloads/

The exact stacktrace I get when running `lein` is:

```
C:\Program Files\Zulu\compiler>lein
Exception in thread "main" java.lang.ExceptionInInitializerError
        at java.lang.Class.forName0(java.base@9-ea/Native Method)
        at java.lang.Class.forName(java.base@9-ea/Class.java:378)
        at clojure.lang.RT.classForName(RT.java:2168)
        at clojure.lang.RT.classForName(RT.java:2177)
        at clojure.lang.RT.loadClassForName(RT.java:2196)
        at clojure.lang.RT.load(RT.java:443)
        at clojure.lang.RT.load(RT.java:419)
        at clojure.core$load$fn__5677.invoke(core.clj:5893)
        at clojure.core$load.invokeStatic(core.clj:5892)
        at clojure.core$load.doInvoke(core.clj:5876)
        at clojure.lang.RestFn.invoke(RestFn.java:408)
        at clojure.core$load_one.invokeStatic(core.clj:5697)
        at clojure.core$load_one.invoke(core.clj:5692)
        at clojure.core$load_lib$fn__5626.invoke(core.clj:5737)
        at clojure.core$load_lib.invokeStatic(core.clj:5736)
        at clojure.core$load_lib.doInvoke(core.clj:5717)
        at clojure.lang.RestFn.applyTo(RestFn.java:142)
        at clojure.core$apply.invokeStatic(core.clj:648)
[...]

PR #2152, opened 2016-05-27
Get lein partially working under Java 9

tobias:
This includes three changes:

* Exclude the dynapath that comes in from bultitude and pomegranate, and
  depend directly on a newer version that operates properly under Java 9
* Ensure we have a modifiable classloader before asking pomegranate to
  modify it. Before Java 9, the AppClassLoader was a URLClassLoader, and
  therefore modifiable. This is no longer the case with Java 9.
* Remove duplicated classpath init code from `eval-in :leiningen`, and
  replace it with a call to `project/init-lein-classpath`. This required
  making the latter function public.

These changes allow some lein functionality to work under Java 9 - there
is a Java 9 issue with data.xml that needs to be fixed before lein's
tests can all pass (http://dev.clojure.org/jira/browse/DXML-32).

Once bultitude and pomegranate are updated to use the newer dynapath, 
the exclusions here can be removed.

This resolves the primary issue in #2149.

Doing this as a PR to get a second set of eyes on it to make sure I'm not inadvertently breaking something.

### DIFF
--- a/leiningen-core/project.clj
+++ b/leiningen-core/project.clj
@@ -4,10 +4,11 @@
             :url "http://www.eclipse.org/legal/epl-v10.html"}
   :description "Library for core functionality of Leiningen."
   :dependencies [[org.clojure/clojure "1.8.0"]
-                 [bultitude "0.2.8"]
+                 [bultitude "0.2.8" :exclusions [org.tcrawley/dynapath]]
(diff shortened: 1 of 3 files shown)


hyPiRion:
This seems straightforward enough, and I've not seen anything strange when I've used it so I presume it's good. Thanks!

Issue #2091, opened 2016-02-16
java.lang.UnsupportedOperationException:


gigafone:
Can you help improve the following error message?

If I make the following mistake and try:

```sh
cd bar # note: there is no `project.clj` in here
lein run -m foo
```

I get:

```sh
clojure.lang.Compiler$CompilerException: java.lang.UnsupportedOperationException: Unknown Collection type, compiling:(NO_SOURCE_PATH:0:0)
 at clojure.lang.Compiler.analyzeSeq (Compiler.java:6875)
    clojure.lang.Compiler.analyze (Compiler.java:6669)
    clojure.lang.Compiler.eval (Compiler.java:6924)
    clojure.lang.Compiler.eval (Compiler.java:6917)
    clojure.lang.Compiler.eval (Compiler.java:6890)
    clojure.core$eval.invokeStatic (core.clj:3105)
    clojure.core$eval.invoke (core.clj:3101)
    leiningen.core.eval$fn__5670.invokeStatic (eval.clj:349)
    leiningen.core.eval/fn (eval.clj:338)
    clojure.lang.MultiFn.invoke (MultiFn.java:233)
    leiningen.core.eval$eval_in_project.invokeStatic (eval.clj:373)
    leiningen.core.eval$eval_in_project.invoke (eval.clj:363)
    leiningen.core.eval$eval_in_project.invokeStatic (eval.clj:367)
    leiningen.core.eval$eval_in_project.invoke (eval.clj:363)
    leiningen.run$run_main.invokeStatic (run.clj:128)
    leiningen.run$run_main.invoke (run.clj:123)
    leiningen.run$run.invokeStatic (run.clj:155)
    leiningen.run$run.doInvoke (run.clj:132)
    clojure.lang.RestFn.invoke (RestFn.java:439)
    clojure.lang.Var.invoke (Var.java:388)
    clojure.lang.AFn.applyToHelper (AFn.java:160)
    clojure.lang.Var.applyTo (Var.java:700)
    clojure.core$apply.invokeStatic (core.clj:648)
    clojure.core$apply.invoke (core.clj:641)
    leiningen.core.main$partial_task$fn__5829.doInvoke (main.clj:272)
    clojure.lang.RestFn.applyTo (RestFn.java:139)
    clojure.lang.AFunction$1.doInvoke (AFunction.java:29)
    clojure.lang.RestFn.applyTo (RestFn.java:137)
    clojure.core$apply.invokeStatic (core.clj:648)
[...]

PR #2157, opened 2016-06-09
Fix for issue #2091

3duard0:
This PR fix issue #2091
`lein run` would crash when evaluating in `:leiningen` instead of `:subprocess`.
This would happen anytime that the `eval` function was called with the `run-form` as the form and no arg was passed.
As shown here https://groups.google.com/forum/#!topic/clojure/7xa7Yg0-htc
`eval` can't handle well a form that contains an evaluated empty lazy-seq, throwing an `java.lang.UnsupportedOperationException`. 
Because of this the form was not being evaluating, not showing an informative error message for the user.

### DIFF
--- a/src/leiningen/run.clj
+++ b/src/leiningen/run.clj
@@ -124,7 +124,9 @@
   "Loads the project namespaces as well as all its dependencies and then calls
   ns/f, passing it the args."
   [project given prep-type args]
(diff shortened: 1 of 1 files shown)


hyPiRion:
Thanks, this seems to be safe and solves the issue nicely :)

Issue #2156, opened 2016-06-06
lein vcs assert-committed is broken


a1exsh:
Preconditions:
```
$ lein -v
Leiningen 2.6.1 on Java 1.8.0_91 OpenJDK 64-Bit Server VM
```

Setup:
```
$ lein new default
Generating a project called default based on the 'default' template.
The default template is intended for library projects, not applications.
To see other templates (app, plugin, etc), try `lein help new`.
$ cd default
```

```
$ git init .
Initialized empty Git repository in /.../default/.git/
$ git add project.clj
$ git commit -m "add project.clj"
```

```
$ $EDITOR project.clj
$ git diff --stat
 project.clj | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)
```

Observation:
```
$ lein vcs assert-committed && echo OK
On branch master
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   project.clj

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	.gitignore
	CHANGELOG.md
	LICENSE
	README.md
	doc/
	src/
	test/

no changes added to commit (use "git add" and/or "git commit -a")
OK
```
Expected: no OK message, `Uncommitted changes in ...` reported.

Observation:
```
$ git add project.clj
$ lein vcs assert-committed && echo OK
On branch master
Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

	modified:   project.clj

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	.gitignore
	CHANGELOG.md
	LICENSE
	README.md
	doc/
	src/
	test/

OK
```
Expected: no OK message, `Uncommitted changes in ...` reported.

3duard0:
Yes, it seems to be broken.

The error was introduced by the fix of error #1857 at this commit: c35a1de4f0b45365a97e6e54c1eba309b45b7f46

The function `eval/sh` is not piping output from `*out*` anymore, instead it is piping from `System/out`.  `with-out-str` when called with `eval/sh` as parameter is returning empty string, thus failing to match `git status` output.

That's all I've found about the problem, I still don't know a fix for this that won't break fix #1857

3duard0:
@hyPiRion I've added a fix for this, but I'm still new to clojure, if anything is not okay I'll be glad to correct.

hyPiRion:
Fixed by #2159.

PR #2159, opened 2016-06-10
Fix for issue #2156

3duard0:
Usage of with-out-str with eval/sh should be replaced by utils/with-system-out-str
as eval/sh is writing output to System/out.


hyPiRion:
Thanks!

PR #2164, opened 2016-06-23
Add support for reading project from Reader instead of file

cprice404:
In certain cases (e.g. reading a project.clj file out of a jar,
rather than from a file on disk) it is useful for the `read`
and `read-raw` functions to support being passed a `Reader` object,
rather than a File or path.

This commit modifies the `read-raw` function to support being
passed a Reader.  It also modifies the `defproject` macro to
avoid the assumption that `*file*` will always have a non-nil
value, because that assumption causes an NPE in the case where
a Reader is the source of the project definition.

### DIFF
--- a/leiningen-core/src/leiningen/core/project.clj
+++ b/leiningen-core/src/leiningen/core/project.clj
@@ -11,7 +11,7 @@
             [leiningen.core.user :as user]
             [leiningen.core.classpath :as classpath])
   (:import (clojure.lang DynamicClassLoader)
-           (java.io PushbackReader)))
+           (java.io PushbackReader Reader)))
 
 (defn make-project-properties [project]
   (with-open [baos (java.io.ByteArrayOutputStream.)]
@@ -400,10 +400,11 @@
   "The project.clj file must either def a project map or call this macro.
   See `lein help sample` to see what arguments it accepts."
   [project-name version & args]
-  `(let [args# ~(unquote-project (argument-list->argument-map args))
-         root# ~(.getParent (io/file *file*))]
-     (def ~'project
-       (make args# '~project-name ~version root#))))
+  (let [f (io/file *file*)]
+    `(let [args# ~(unquote-project (argument-list->argument-map args))
+           root# ~(if f (.getParent f))]
+       (def ~'project
+         (make args# '~project-name ~version root#)))))
 
 (defn- add-exclusions [exclusions dep]
   (dependency-vec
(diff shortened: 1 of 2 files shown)


cprice404:
This is related to #2067 , in that in order to implement the 'lein parent' functionality that pulls in a parent project file from a set of maven coordinates, we need to be able to read projects from sources other than files on disk (e.g. from a stream coming out of a zip/jar file entry).  This is necessary regardless of whether or not we decide to implement the parent functionality as a plugin or as part of the core.

cprice404:
@senior this is the other PR that could use a review from you while you're poking at all of this stuff.

leiningen-core/src/leiningen/core/project.clj
```diff
@@ -947,15 +948,18 @@ Also merges default profiles."
...
   "Read project file without loading certificates, plugins, middleware, etc."
-  [file]
+  [source]
   (locking read-raw
     (binding [*ns* (find-ns 'leiningen.core.project)]
-      (try (load-file file)
-           (catch Exception e
-             (throw (Exception. (format "Error loading %s" file) e)))))
+      (try
+        (if (instance? Reader source)
```
senior: Any reason not to use something like clojure.io/reader to coerce to a reader, either always or with an inverted logic.

cprice404: That sounds sane.  Let me try it.

cprice404: When I make that change it gives me an error:

```
>  bin/lein test
Couldn't find project.clj, which is needed for test
```

I remembered why this is.  It's because [the `defproject` macro relies on the presence of the dynamic variable `*file*`](https://github.com/cprice404/leiningen/blob/2a30d862f6bbf2a9633836fdc4b72ab6c2f623d6/leiningen-core/src/leiningen/core/project.clj#L403) to set the `:root` key in the project map.  This var exists if you use clojure's `load-file` macro, but not if you use `load-reader`.

There are a handful of places in the lein code (e.g. when you are running basic commands like `test` from the command line) that rely on `:root` being set properly.  This is not necessary for our use case of reading the project file from the `lein-parent` plugin.

So, the current code was the cleanest way that I came up with to allow a project file to be loaded from a maven artifact (w/o actually existing on disk), while not breaking the other use cases where you are interacting with the project file from `lein` directly.

It's possible there's another way to handle this, but I think it would involve more drastic changes than what I have here, and might require some arity-overloads to some of these functions in order to avoid breaking backward compat.  Open to suggestions, though.

senior: Makes sense. Sounds like your original solution is best here.

senior:
Looks good to me

pcarlisle:
Looks good

cprice404:
@hyPiRion I'm going to go ahead and merge this as it's the last piece of the puzzle for the `lein-parent` stuff that we are trying to play with, and I've gotten review from the co-workers who I think would have the most insight into these changes.  I'll definitely keep an eye out in case it breaks anything.

Would love to get a lein 2.7.0 release out sometime that contains these changes.  If there's anything I can do to help with that, or if there is a list of other tickets that you were hoping to get some attention on before you cut a new release, let me know and I will be happy to try to help.

Thanks!

Issue #1782, opened 2014-12-08
Reconciling timing of project prepping and test namespace discovery


cemerick:
If a task added to `:prep-tasks` generates sources containing tests, those test namespaces will not be run because `leiningen.test` performs its namespace discovery before `eval-in-project` (and therefore `:prep-tasks` tasks) is run.

A real-world example of this is cljx (described more completely with a workaround [here](https://github.com/lynaghk/cljx/issues/54)), which is moving away from hooks to use `:prep-tasks` (as advised [here](https://github.com/technomancy/leiningen/issues/1586)).

How can this be reconciled?

hyPiRion:
Fixed by #2165.

PR #2165, opened 2016-06-24
Add an explicit prep call before the test namespaces are calculated.

martinreck:
https://github.com/technomancy/leiningen/issues/1782


hyPiRion:
Thanks :)

PR #2169, opened 2016-07-11
Testing template doc

chris-etheridge:
I could not find much information about testing a local copy of a template in the documentation 😄

### DIFF
--- a/doc/TEMPLATES.md
+++ b/doc/TEMPLATES.md
@@ -129,6 +129,21 @@ change the mustache delimiter temporarily, like so:
 <%={{ }}=%>
 ```
 
+## Testing your template
+
+Templates can be tested locally with ease. 
+
+First, make sure you're in your template project directory, then:
+
+```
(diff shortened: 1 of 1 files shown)


cprice404:
Hey @chris-etheridge - thanks for the PR!  There were a few sentences in the existing docs that mention how to test a template, but I think you were right that they weren't called out as clearly as they could have been.  I made a section header for them and put in a bit more info here:

https://github.com/technomancy/leiningen/commit/7354e25e23290891837198604c93fd9cf2d16c63

Does that look ok to you?  I don't think you should need to run `lein jar` or change to the `target` directory in order to run your template.  I'll close this PR for now in hopes that the change linked above is sufficient, but if for some reason those steps don't work for you, definitely let me know and we can re-open this and get to the bottom of it.

chris-etheridge:
Yup that looks totally cool! Thanks for adding this in.

PR #2171, opened 2016-07-12
Update test for clojars cert expiration date - fixes #2170

### LINKED ISSUE #2170: Tests failing related to clojars-ssl-cert
After the most recently merged PR, [travis is failing](https://travis-ci.org/technomancy/leiningen/jobs/144248991) with the following error:

```
lein test :only leiningen.test.keys/clojars-ssl-cert

FAIL in (clojars-ssl-cert) (keys.clj:13)
If this fails, yell at _ato to get a new SSL cert.
expected: (< (System/currentTimeMillis) (.getTime month-before-cert-expiry))
  actual: (not (< 1468349427928 1468281600000))
```

That sounds like the clojars server cert may be expired?  Which could theoretically explain some of the other SSL-related issues people have been running into lately?

I will attempt to look into it further but if anyone knows more about this error than I do, info would be appreciated!

cprice404:


cprice404:
travis isn't quite done but one cell went green, and I have to run to a meeting, and don't want to forget to merge this :)

PR #2174, opened 2016-07-16
New Powershell bootstrap

brianary:
This is a very literal port of lein.bat to PowerShell. No more screens of errors or usage messages at each launch! 

TODO:
    - Determine which (if any) environment variables are used from within Leiningen/Clojure
      and convert the rest to local- or script-scoped variables.
    - Probably should parse the version from the newest .jar filename, rather than hard-code
      it and search for the .jar file from it.
    - Further reduce/simplify/refactor away from batch-file idioms toward idiomatic PowerShell.


winks:
Hey, this looks great!

Just wanted to test it and Win7 told me to "get-help about_signing" because apparently I haven't allowed executing random PowerShell scripts? That was invoking `lein.cmd`.

I'm a total noob with PS, I just did some debugging when we made the change to try to download the JAR in `lein.bat` via PowerShell.

So, any experience or hints if this restriction is typical for Windows installations or if I locked somethign down in the past? Because that would probably limit the usefulness quite a lot.

But of course I'm gonna read up now and maybe try on another box.

brianary:
Thanks! Yes, I think most versions of Windows default the Execution Policy
to Restricted. I can add a parameter ("-ExecutionPolicy Unrestricted") to
the script invocation in lein.cmd to mitigate that when I get home.

winks:
That helped, thanks.

been playing around a bit now, everything seems to work (or fail in the same way as lein.bat (2.6.1) does).

hyPiRion:
Thanks for the work on this! Since I'm no good at neither batch or powershell, what's the status here? Could we in theory replace the batch script with the PS variant?

brianary:
Yes, there are probably more optimizations that can be done (e.g. are all
those environment variables used inside the JVM?), but theoretically
lein.cmd should be able to replace lein.bat.

hyPiRion:
Excellent – I'll try it out and see if we're able to move over to the script during 3.7.1; I'd like to wait one release so that people can test the waters first.

brianary:
Sounds good!

hyPiRion:
So there were two small quoting issues here, but otherwise this seems fine:

```diff
$JavaArgs = @(
    '-client',$env:LEIN_JVM_OPTS,
-    "-Dclojure.compile.path=$PWD/target/classes", #TODO: is this what lein.bat was going for?
-    "-Dleiningen.original.pwd=`"$env:ORIGINAL_PWD`"",
+    "`"-Dclojure.compile.path=$PWD/target/classes`"", #TODO: Only add these lines when initializing from source
+    "`"-Dleiningen.original.pwd=$env:ORIGINAL_PWD`"",
    '-cp',$env:CLASSPATH,
    'clojure.main',
    '-m','leiningen.core.main'
)
```

Does that seem right? This made folders with spaces work at least, without those changes it didn't.

Also, both `lein.ps1` and `lein.cmd` does not contain the usual windows CRLF line endings. Should we keep newlines for windows CRLF or is just LF fine these days?

hyPiRion:
Also, final question: Users need to put both `lein.cmd` and `lein.ps1` in the same directory for this to work, right?

brianary:
That quote fix looks good. Maybe my .gitconfig ate the CR, but I'd force
CRLF if possible on those files. Those files do need to be in the same
directory the way it's currently written, yes.

hyPiRion:
Alright, thanks for the pointers. I'll fix the remainding issues and get the word out with the next release :)

brianary:
Great! If you run side-by-side for a while, people can use lein.cmd, and
fall back to lein.bat if needed.

PR #2179, opened 2016-07-28
Parse GitHub usernames with - in them

danielcompton:
GitHub usernames only allow alphanumeric characters and single hypens. This commit fixes the regex to allow dashes, and remove _ as _ aren't valid in Github usernames.

screenshot of safari technology preview 29-07-16 10-14-14 am


hyPiRion:
Nice catch; thanks!

PR #2180, opened 2016-07-31
Fixed adding %LEIN_JAR% to %CLASSPATH% (preserving original %CLASSPATH% values)

bcalco:
Without this lein can't find the JAR in the %USERHOME%/.lein/self-installed directory when %USERHOME% has spaces in it. Note, I add %CLASSPATH% to prevent any previously set values in %CLASSPATH% from getting whacked.

### DIFF
--- a/bin/lein.bat
+++ b/bin/lein.bat
@@ -115,7 +115,7 @@ rem the paths inside the bootstrap file do not already contain double quotes but
 
     :: Not running from a checkout.
     if not exist "%LEIN_JAR%" goto NO_LEIN_JAR
-    set CLASSPATH=%LEIN_JAR%
+    set CLASSPATH="%LEIN_JAR%;%CLASSPATH%"
   
     if exist ".lein-classpath" (
         for /f "tokens=* delims= " %%i in (.lein-classpath) do (


winks:
If you look at commit 9750e0ebb96f7c296996705deee17f2c2c6a8497 there's a comment that reads:

> +	rem CLASSPATH must contain each path element double quoted and separated by semicolons ie. "c:\some folder";"c:\some other folder";"c:\semicolons;;;;in;name;work okay"
> +	rem (no end semicolon required)

hyPiRion:
I'm pretty sure I merged in another which (un?)quoted another part further down the script to fix this. I'll attempt to check if this is still an issue on master, and if it is, I'll merge it in. Happy to hear from people running Windows on whether the lein.bat on master works fine with spaces in usernames.

hyPiRion:
So I tested stuff around and indeed, a username with spaces in it will fail with stable. I checked out master and it worked fine there, but patching in this caused it to fail again.

I'm not sure if the original classpath has to be appended, as we're only running lein here. I would be surprised if that one leaks out of the batch script, but if that's the case or you have another good reason for keeping it in, feel free to patch this PR so that it plays fine with the master version of lein.bat and comment here.

Either way, thanks for the fix, even though it seems to already be fixed :)

Issue #2181, opened 2016-08-03
Tutorial typo causes compilation error


webappzero:
Example file path `src/my_stuff.clj` in line 558  of doc/TUTORIAL.md caused compilation error upon execution of `lein uberjar`.

Changing path to `src/my/stuff.clj` fixed issue on my system.

webappzero:
Not sure if this is the right fix?

webappzero:
The error was mine and not the Tutorial's. Please ignore.

PR #2182, opened 2016-08-03
Fix for issue #2181

webappzero:
Edit file path to match the namespace in the code to fix compilation error in Tutorial example.


webappzero:
Fixes issue #2181

webappzero:
Not sure this fixes issue.

cprice404:
hrm.  `src/my_stuff/core.clj` seems like the correct path to me, if the namespace inside the clojure file is `my-stuff.core` (which is what it looks like in the examples).  If you were using a namespace of `my.stuff.core`, then the path you use here (`src/my/stuff/core`) would be correct, but it looks to me like the example says `ns my-stuff.core`.

webappzero:
Oops. You and the Tutorial are both correct. Following the instructions more closely I see where I made a mistake. Thank you!

PR #2206, opened 2016-09-16
Fix checking for SNAPSHOT deps with managed deps - fixes #2205

### LINKED ISSUE #2205: NPE when doing an install/deploy of a non-SNAPSHOT version of a project, when using managed dependencies
Apparently I did all of my testing around the managed dependencies stuff only using SNAPSHOT versions of the various artifacts.  :(

Today we tried to do a `lein install` on a non-SNAPSHOT version of a project, which was using lein-parent/managed-dependencies to inherit dependency version numbers.  So, the project has some deps that don't include versions, like so:

```clj
:dependencies [[org.clojure/clojure]
                 [org.clojure/tools.logging]
                 [org.clojure/tools.cli]

                 [clj-time]]
```

When you remove the `SNAPSHOT` version from the project, it triggers this line of code:
https://github.com/technomancy/leiningen/blob/33eca8dba1bc9f01f94e21f7053702955a73b23a/src/leiningen/pom.clj#L370

```clj
(defn check-for-snapshot-deps [project]
  (when (and (not (snapshot? project))
             (not (System/getenv "LEIN_SNAPSHOTS_IN_RELEASE"))
             (some #(re-find #"SNAPSHOT" (second %)) (:dependencies project)))
    (main/abort "Release versions may not depend upon snapshots."
                "\nFreeze snapshots to dated versions or set the"
"LEIN_SNAPSHOTS_IN_RELEASE environment variable to override.")))
```

The `(second %)` causes the NPE on these deps.

I need to file a patch to have this code do the resolution of the final version numbers from the managed-dependencies prior to executing this line of code.  Will try to have a PR up tomorrow.

cprice404:
This commit fixes #2205: when using managed dependencies in
a project, if the project itself was a non-SNAPSHOT version,
then when `lein install` and similar tasks were executed,
lein would check to make sure that none of the dependencies
were a SNAPSHOT version.  lein was checking that directly against
the `:dependencies` vector, which would cause an NPE if a dependency
didn't specify a version number (because it was inheriting it from
the `:managed-dependencies` section).

This commit fixes the bug by calling the code that merges the
`:dependencies` and `:managed-dependencies` vectors together prior
to performing the SNAPSHOT check.

### DIFF
--- a/src/leiningen/pom.clj
+++ b/src/leiningen/pom.clj
@@ -7,7 +7,8 @@
             [clojure.set :as set]
             [clojure.string :as s]
             [clojure.java.shell :as sh]
-            [clojure.data.xml :as xml]))
+            [clojure.data.xml :as xml]
+            [leiningen.core.classpath :as classpath]))
 
 (defn- relativize [project]
   (let [root (str (:root project) (System/getProperty "file.separator"))]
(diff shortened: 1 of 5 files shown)


cprice404:
There is a travis failure here on JDK8, on a 'bad-aot' test, which I wouldn't think my PR would have anything to do with... but, investigating.

cprice404:
The test passed when re-run, so I guess there is just some kind of transient failure case in that test?

hyPiRion:
@cprice404: Yeah, JDK8 fails every now and then. If there's no obvious reason why a JDK version fails when the others run just fine, I just restart it and it almost always passes then.

There are some other cases where dependencies fail to download, in which case I restart as well.

hyPiRion:
Looks good; thanks!

PR #2207, opened 2016-09-16
Don't require `nil` for version in managed deps - fixes #2195

### LINKED ISSUE #2195: managed dependencies: look into supporting `:exclusions`, etc., without requiring explicit `nil` for version number
As discussed in #2194 , the syntax for using managed dependencies in cases where you need additional modifiers for a dependency (e.g. `:classifier`, `:exclusions`) is not entirely intuitive.

If you don't need any such modifiers, you can use managed dependencies by just omitting the version string from your declared dependency:

```clj
:dependencies [[clj-time]]
```

You can also explicitly specify a `nil` for the version string, which has the same effect but is not entirely intuitive:

```clj
:dependencies [[clj-time nil]]
```

If you do need modifiers, then only the latter syntax works:

```clj
:dependencies [[clj-time nil :exclusions [foo]]
```

This syntax, which is probably more intuitive to users, does *not* currently work:

```clj
:dependencies [[clj-time :exclusions [foo]]
```

It would definitely be nice to be able to support this last syntax.  My recollection was that doing so was going to require some more intrusive changes to the lein dependency code, so I didn't try to bite that off in the first pass, but it is probably worth revisiting.

If my memory serves, this would probably also require some changes to the upstream `pomegranate` library.

cprice404:
Prior to this commit, if you wanted to use modifiers such as
`:exclusions` or `:classifier` for a dependency whose version you
were managing with `:managed-dependencies`, you would need to
explicitly pass a `nil` as the version string in the dependency
tuple.  This commit adds some logic to coerce the vectors before
they are processed, so that if the version string is simply
omitted instead of being set to `nil`, the `nil` will be implicitly
inserted and things will continue to work as before.

This provides a slightly nicer and more intuitive UX for the
managed-dependencies feature.

### DIFF
--- a/leiningen-core/src/leiningen/core/classpath.clj
+++ b/leiningen-core/src/leiningen/core/classpath.clj
@@ -548,7 +548,7 @@
       ;; it's important to preserve the metadata, because it is used for
       ;; profile merging, etc.
       (with-meta
-       (concat [id version] opts)
+       (into [id version] opts)
        (meta dep)))))
 
 (defn normalize-dep-vectors
--- a/leiningen-core/src/leiningen/core/classpath.clj
+++ b/leiningen-core/src/leiningen/core/classpath.clj
@@ -531,6 +531,36 @@
                                    :managed-dependencies)]
     (apply resolve-managed-dependencies dependencies-key managed-dependencies-key project rest)))
 
+(defn normalize-dep-vector
+  "Normalize the vector for a single dependency, to ensure it is compatible with
+  the format expected by pomegranate.  The main purpose of this function is to
+  to detect the case where the version string for a dependency has been omitted,
+  due to the use of `:managed-dependencies`, and to inject a `nil` into the
+  vector in the place where the version string should be."
+  [dep]
+  (if dep
+    (let [id (first dep)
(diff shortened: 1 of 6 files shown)


cprice404:
Rebased now that #2206 is merged.

leiningen-core/src/leiningen/core/classpath.clj
```diff
@@ -531,14 +531,46 @@
...
+    (let [id (first dep)
+          sec (second dep)
+          version (if-not (keyword? sec) sec)
+          opts (if (keyword? sec)
+                 (nthrest dep 1)
+                 (nthrest dep 2))]
+      ;; it's important to preserve the metadata, because it is used for
+      ;; profile merging, etc.
+      (with-meta
+       (concat [id version] opts)
```
hyPiRion: I'd _maybe_ use `(into [id version] opts)` instead of concat, because pomegranate consistently talks about vectors as input. But it's splitting hairs if pomegranate doesn't care.

cprice404: Makes sense - added a commit that does that.

PR #2211, opened 2016-09-22
Powershell fixes

scottdw:
The first commit fixes an error that was being thrown when using a proxy:
`Cannot convert the "System.Collections.Hashtable" value of type "System.Collections.Hashtable" to type
"System.Management.Automation.PSBoundParametersDictionary".`

The second I'm unable to test but appears more correct than what was there before:
1) The pending and current powershell scripts were reversed in the Update-Self function.
2) The Install-Self function was being called for the current version not the new one.

### DIFF
--- a/bin/lein.ps1
+++ b/bin/lein.ps1
@@ -44,10 +44,8 @@ function Initialize-Environment
     {
         $proxy = [Net.WebRequest]::DefaultWebProxy.GetProxy('https://github.com/')
         Write-Verbose "Using proxy: $proxy"
-        $Script:PSBoundParameters = @{
-            'Invoke-WebRequest:Proxy' = $proxy
-            'Invoke-WebRequest:ProxyUseDefaultCredentials' = $true
-        }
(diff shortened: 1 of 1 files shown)


scottdw:
Is there any more information I can provide to ensure these fixes go into the next release?

hyPiRion:
@scottdw: Sorry – I usually test this through a VM at home, but went out of memory and had wipe that one for some time. I have a new disk now so I'll see if I can check this during this weekend.

marco-m:
Hello, any news ?

technomancy:
@marco-m this is just waiting for someone who knows about powershell to confirm that it works and is a good idea.

scottdw:
The first commit fixes a runtime exception which I assume only affects certain versions of powershell.

The second is verifiable by inspection, the original script ran the install function of the current script which had the old version number set. Then deleted the just downloaded file and moved the existing file but only after the current powershell session was exited.

On the other hand the bat file works fine in powershell :).

marco-m:
thanks both for the quick reply :-)

PR #2213, opened 2016-09-27
Tolerate keywords for version string - fixes #2212

### LINKED ISSUE #2212: classifier option fails in 2.7.1
Works in 2.6.1
Works in 2.7.0
Fails in 2.7.1
Has anyone experienced this?

cprice404:
This commit modifies the logic for normalizing the dependency
vectors.  The previous implementation would not tolerate a keyword
for the `version` identifier in the second position in the tuple.
This caused problems with the lein-module plugin, which allows
keywords to be used as the version identifier.

This implementation takes a different approach, and only assumes
that the vector must end with keyword/value option pairs (if any
options are specified).  It extracts those first, and then tolerates
any remaining value to be interpreted as the version identifier
(including keywords).

This commit also adds some more strict checking of the contents
of the dependency vectors; specifically, it will throw an exception
if the options at the end of the vector are not keyword/value pairs,
or if the artifact id in the first position in the dependency tuple
is not a symbol.

### DIFF
--- a/leiningen-core/src/leiningen/core/classpath.clj
+++ b/leiningen-core/src/leiningen/core/classpath.clj
@@ -531,25 +531,44 @@
                                    :managed-dependencies)]
     (apply resolve-managed-dependencies dependencies-key managed-dependencies-key project rest)))
 
+(defn- extract-dep-options
+  "Get 'options' values from the end of a dependency vector.  This basically just pulls off pairs
+  of values from the end of the vector and then returns a tuple whose first value is those pairs
+  (re-assembled into a single vector), and whose second value is whatever is left from the original
+  dependency vector after all pairs have been removed."
+  [acc dep-rest]
+  (if (<= (count dep-rest) 1)
+    [acc dep-rest]
+    (extract-dep-options
+      (concat (take-last 2 dep-rest) acc)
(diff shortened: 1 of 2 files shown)


cprice404:
@hyPiRion curious to hear what you think of this approach.  I don't love it, but didn't have a better idea... we've unfortunately moved into the land of heuristics here :(

Also, curious to see if you think that the additional error checking that I added is valuable, or if you can think of other situations where people might be doing fancy things with the dep vector.  I don't mind removing that validation code if it doesn't seem like a good idea to you.

Would be really nice to have a clojure.spec specification for those dependency vectors :/

hyPiRion:
I think better error validation is great, but I think it's orthogonal to this particular issue (if possible it should be done right after plugins have ran their middleware). I think Bruce Hauman has worked on good error messages via spec, so we should definitely look into that.

The approach seems fine for me, I just have a small question: From what I can see by that code, it effectively does a check whether the vector is odd or not, and if it is, inject a `nil` where the version is supposed to be. Is this right or have I missed something? I made #2214 either way, and it _seems_ to work exactly the same, but would be happy to get your thoughts on that approach.

cprice404:
Closing this in favor of #2214 .

PR #2214, opened 2016-09-28
Allow keyword dependency versions, fixes #2212

### LINKED ISSUE #2212: classifier option fails in 2.7.1

hyPiRion:
This is an alternative to #2213 and also solves #2212, albeit with a different solution:

The assumption is that dependency vectors have always been even elements long until we added managed dependencies (Otherwise they'd blow up at runtime). If that's the case, then the only vectors of odd length are the ones without a version explicitly set. This means we can just check a vector's length to verify whether we should inject a `nil` or not after the version identifier.

### DIFF
--- a/leiningen-core/src/leiningen/core/classpath.clj
+++ b/leiningen-core/src/leiningen/core/classpath.clj
@@ -538,17 +538,17 @@
   due to the use of `:managed-dependencies`, and to inject a `nil` into the
   vector in the place where the version string should be."
   [dep]
-  (if dep
+  ;; Some plugins may replace a keyword with a version string later on, so
+  ;; assume that even lenght vectors are alright. If not, then they will blow up
+  ;; at a later stage.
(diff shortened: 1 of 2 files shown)


leiningen-core/src/leiningen/core/classpath.clj
```diff
@@ -538,17 +538,17 @@
   due to the use of `:managed-dependencies`, and to inject a `nil` into the
   vector in the place where the version string should be."
   [dep]
-  (if dep
+  ;; Some plugins may replace a keyword with a version string later on, so
+  ;; assume that even lenght vectors are alright. If not, then they will blow up
```
cprice404: typo in `lenght`.

cprice404:
I think you're right, and this is simpler.  :+1:

Issue #2216, opened 2016-10-10
merged profiles do not resolve managed dependencies properly


cprice404:
I'm seeing a case where, when I have a project that uses managed dependencies, and a profile where we specify `:dependencies` with a `^:replace`, somehow the code that checks whether the dependencies have version numbers or not and fills in a `nil` if they don't is not getting called.  This causes an `IndexOutOfBounds` exception on this line, in `get-original-dependency`:

https://github.com/technomancy/leiningen/blob/2.7.1/leiningen-core/src/leiningen/core/classpath.clj#L434

because `(subvec v 0 2)` gets called on a vector that may only have one entry.

There are definitely other code paths that hit this line of code, and the input consistently has the `nil` values inserted in as the second value of the tuples in the input.  So it has something to do with profiles being merged.  Not sure yet whether or not it only happens with `:replace`; I'll work up a reproducer and see if I can file a PR.

Note that it is possible to work around this issue for now by explicitly adding a `nil` in as the second item in the dependency tuples in the profile in the `project.clj` file.

glts:
This came with a pull request #2217 which was merged long ago. We can close this.

PR #2217, opened 2016-10-11
(#2216) Fix bug with managed deps in profiles w/:replace

cprice404:
Prior to this commit, profiles with `^:replace` on the dependencies
list would never end up having their dependencies vector normalized
so that it would have `nil` placeholders for the versions of
dependencies that were inheriting their version from `:managed-dependencies`.

This commit normalizes the dependencies vector of a profile during
initialization, to make sure that it will always be normalized.


cprice404:
This should fix #2216

hyPiRion:
Looks good, thanks :)

PR #2219, opened 2016-10-20
Tweaks the README

AdilKhn:
- adds permissions label for the leiningen executable


hyPiRion:
This would be fine to include, but I think just adding the permission name would cause it to be even more cryptic. Perhaps if we add in something like 

> 4. Set it to be executable. (`chmod 755 ~/bin/lein`, 755 is octal for the permissions rwx-r-x-r-x)

it would be more readable.

That being said, an even better solution would probably be to just change the chmod call to `chmod +x ~/bin/lein`, which should be more readable.

technomancy:
Fixed in a441ba1.

PR #2220, opened 2016-11-03
Infer classifier when deploying adhoc files

### LINKED ISSUE #2161: lein deploy doesn't support :classifier parameter
From lein help, the arguments for lein deploy are:

Arguments: ([] [repository] [repository identifier version & files])

It looks like there's no way to deploy a package with both a version and a :classifier keyword, as in this example from https://github.com/technomancy/leiningen/blob/master/sample.project.clj:

:dependencies [[org.jclouds/jclouds "1.0" :classifier "jdk15"]]

I worked around this by deploying with mvn, but it would be more convenient to just use lein deploy.

rymndhng:
Infer the classifier from the file path if it exists when performing `lein deploy`. 

Fixes #2161 

See: http://maven.apache.org/plugins/maven-deploy-plugin/examples/deploying-with-classifiers.html

### DIFF
--- a/src/leiningen/deploy.clj
+++ b/src/leiningen/deploy.clj
@@ -155,6 +155,16 @@
       "pom"
       (last (.split f "\\.")))))
 
+(defn classifier
+  "The classifier is be located between the version and extension name of the artifact.
+
+  See http://maven.apache.org/plugins/maven-deploy-plugin/examples/deploying-with-classifiers.html "
+  [version f]
+  (let [pattern (re-pattern (format "%s-(.*)\\.%s" version (extension f)))
+        [_ classifier-of] (re-find pattern f)]
+    (when-not (empty? classifier-of)
+      classifier-of)))
+
 (defn- fail-on-empty-project [project]
   (when-not (:root project)
     (main/abort "Couldn't find project.clj, which is needed for deploy task")))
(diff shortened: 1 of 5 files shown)


cprice404:
This looks good to me - might be worth adding a test that actually calls `deploy`, though?

rymndhng:
@cprice404 I added a test. I hope this is sufficient.

hyPiRion:
Yeah, this looks fine – one thing I'd like (but is not very important!) is replacement of regex with basic string functions; I think it should be possible to extract the classifier with just `clojure.string/ends-with?` and `starts-with?` combined with `subs`.

I'll merge it in later today unless @cprice404 beats me to it :)

rymndhng:
@hyPiRion I'm not seeing the solution that you are suggesting. I'm all for it, but I am scratching my head a little.

With this pattern: `{artifact_name}-{version}-{classifier?}.{extension}`, I don't think we can assume `{artifact_name}` is required to be the same as the project name, which makes me hesitant to use `starts-with?`. 

I'd love to have your thoughts though.

hyPiRion:
@rymndhng: Ah, nevermind, I thought the `re-find` call was a `re-matches`.

rymndhng:
@hyPiRion alright 😄  🏁

PR #2225, opened 2016-12-09
Use new clojars repo url

tobias:
This switches the default repo url (for resolving artifacts, not
deploying) to point to the CDN-fronted repo. Note that this repo won't
work with Java 6 - users of 6 will need to manually override the default
url to point to the old one (https://clojars.org/repo).


hyPiRion:
Huh, how come Java 6 doesn't work with this new endpoint?

tobias:
Interesting. It was my understanding that the Fastly CDN doesn't have the proper SSL settings to support Java 6, but either I'm wrong, or the tests don't use the new repo url. I'll take a look.

tobias:
Forget what I said earlier. The new url works just fine with Java 6.

PR #2227, opened 2016-12-12
fix :plugins not use the configuration in the ~/.lein/profiles.clj

gentoor:
The configuration in ~/.lein/profiles.clj will lose efficacy,  when you
write  a top level :plugins in the project.clj.
simple test passed!


technomancy:
Can you please describe the motivation for this change? I can't understand why it's needed.

technomancy:
If you need the `:plugins` entry in `project.clj` to override user level `:plugins`, you can use metadata for that.

PR #2234, opened 2017-01-15
Bump org.apache.maven.indexer version to latest

ehashman:
Rationale: I want to help get leiningen packaged for Debian again ([it hasn't been since 1.7.1](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=819811)). This particular dependency, in addition to the old dynapath and commons-io versions in 2.7.1, are behind the versions currently available in Debian unstable.

At the same time, I'm going to work to get all of the missing dependencies for leiningen in Debian uploaded, and all the out-of-date dependencies in Debian updated.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -11,7 +11,7 @@
                  [commons-io "2.5"]
                  [bultitude "0.2.8"]
                  [stencil "0.5.0" :exclusions [org.clojure/core.cache]]
-                 [org.apache.maven.indexer/indexer-core "4.1.3"
+                 [org.apache.maven.indexer/indexer-core "5.1.1"
                   :exclusions [org.apache.maven/maven-model
                                org.sonatype.aether/aether-api
                                org.sonatype.aether/aether-util


technomancy:
Thanks for taking this on.

Before we bump the version here, I would like to discuss the option of removing the `search` task from Leiningen. When it was created, it made sense to download indices for local search, since they were only around 60MB. Since then they have grown to the gigabyte range, and the servers they are hosted on are quite busy, leading to poor download performance. If you do a web search for maven central index size, you'll see most of the results are people complaining that the indices offer such poor performance as to basically render them unusable.

Central and Clojars both offer excellent search on their respective web interfaces, and most developers would be much better off using those instead. I would submit that the best thing to do with the search task is to replace it with a notice that tells people to perform the search in their browser instead.

If people really feel the need to have offline access to searches, we could either spin it out into an external plugin or remove the maven indexer deps from lein's own `project.clj` and resolve them at runtime when running the `search` task if the user insists that he doesn't want to use the online indices.

It's technically a semver violation, but I think that ship has sailed since the situation with the index size has changed so dramatically since it was introduced, and we don't have any control over that. If it makes it easier for Leiningen to be packaged as part of Debian, I would say it's absolutely worth it.

How do people feel about this?

tobias:
Another option is to replace the index download with a call to the search APIs for Central/Clojars (https://search.maven.org/#api https://github.com/clojars/clojars-web/wiki/Data#json-search-results), though the format of the results are different, so each would require custom parsers.

winks:
+1 for splitting the current search (with downloading an index) out into a plugin
+1 for making `lein search` just do a web request to clojars and/or central

technomancy:
Opened #2235 which removes the need for maven-indexer.

ehashman:
From the peanut gallery: I am totally +1 to removing the dependency. I just submitted this PR because it seemed like the lowest hanging fruit :)

Feel free to close this once #2235 is merged.

technomancy:
Thanks @ehashman. I'm excited to see your packaging work move forward, and please let me know if there are any other changes we can make to smooth things along for you.

technomancy:
@ehashman I noticed you are associated with OpenHatch--I would be interested in feedback if you run into anything that could be improved in how we run this project, especially around how to make it more friendly to first-time contributors. Thanks!

ehashman:
@technomancy I'm pretty busy this week but when I get a chance I'd be happy to chat about that---you can find on me on IRC or the Clojurians Slack :)

technomancy:
I haven't used Slack, but please do come by `#leiningen` on Freenode any time.

PR #2235, opened 2017-01-18
Replace lein search with live API-hitting task.

technomancy:
Addresses the discussion in #2234.

No pagination yet.

Also checks repositories based on the presence of "clojars" or
"central" in the :repositories map; you could theoretically keep these
under a different name or something and it could get confused, but
... why would you?


src/leiningen/search.clj
```diff
@@ -5,231 +5,45 @@
...
-                               query-parts)]
-        (println (format "Searching over %s..."
-                         (string/join ", " search-fields))))
-      (print-results response page))))
+            [clj-http.client :as http]
+            [cheshire.core :as json]))
+
+(defn search-central [query]
+  (println "Searching Central...")
+  (let [url (str "http://search.maven.org/solrsearch/select?wt=json&q=" query)]
```
winks: Why not https by default? typo?

src/leiningen/search.clj
```diff
@@ -5,231 +5,45 @@
...
+    (when (repos "central")
+      (try (search-central query)
+           (catch clojure.lang.ExceptionInfo e
+             (if (= 400 (:status (ex-data e)))
+               (println "Query syntax unsupported by Central.")
+               (throw e)))))
+    (when (repos "clojars")
+      (try (search-clojars query)
+           (catch clojure.lang.ExceptionInfo e
+             (if (= 400 (:status (ex-data e)))
```
winks: Was about to comment that I've been bitten by too many uncaught 40x in my life..
But clojars seems to be strict about 400: https://github.com/clojars/clojars-web/blob/c992f80b73628648e58a9da1a053a251f1325dbf/test/clojars/test/integration/search.clj

Still, any special reason to not catch 4xx and 5xx?

winks:
:1st_place_medal: time-to-market record for this feature :+1:

technomancy:
@winks thanks; I ended up re-throwing non-clj-http errors and separating 400s vs non-400. Missing `https` was in fact a typo.

PR #2238, opened 2017-01-25
Remove lein-newnew references from templates docs

gfredericks:
I don't know for suresies that this makes sense.

My cursory understanding is that lein-newnew was integrated into lein 2, and it seems this doc file was adapted from lein-newnew's docs and had some leftover references.

Hopefully somebody more familiar with the templates than myself can verify that these changes make sense.

It's hard to talk clearly about template templates.

### DIFF
--- a/doc/TEMPLATES.md
+++ b/doc/TEMPLATES.md
@@ -50,8 +50,8 @@ an artifact-id of "lein-template".
 ## Structure
 
 The files that your template will provide to users are in
-`src/leiningen/new/liquid_cool`. lein-newnew starts you off with just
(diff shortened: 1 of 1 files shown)


gfredericks:
That build failure cannot possibly be my fault.

technomancy:
> Caused by: org.sonatype.aether.transfer.ArtifactTransferException: Could not transfer artifact lein-maven:lein-maven:jar:0.1.0 from/to clojars (https://repo.clojars.org/): Read timed out

I think you're in the clear; haha.

technomancy:
Thanks.

Issue #2241, opened 2017-02-01
Leiningen can't install maven artifacts whose names are invalid symbols


cap10morgan:
For example: Specifying `net.3scale/3scale-api` in the dependencies causes lein to throw an invalid token exception when it tries to load the `project.clj`.

cap10morgan:
Specifying this dep like this is a workaround: `[~(symbol "net.3scale" "3scale-api") "3.0.2"]`. So if that's the official / best way to do it, I'm happy to close this.

hyPiRion:
It's currently the best – if not the only – way to do it, yeah.

I think this would happen so infrequently (symbol name must start with digit) that I myself wouldn't look into making this more convenient than it currently is. That being said, I wouldn't mind a patch which allows us to use `"net.3scale/3scale-api"` if it doesn't require too big changes in leiningen core.

cap10morgan:
@hyPiRion Let me know what you think of this PR: #2242.

hyPiRion:
Looks good, thanks! :)

PR #2242, opened 2017-02-02
Add support for string deps in project.clj

cap10morgan:
This is needed for rare cases where maven artifact-ids and/or group-ids
aren't valid symbols (e.g. they start with a number).

Test coverage added.

This PR was suggested in my bug report #2241.

### DIFF
--- a/leiningen-core/pom.xml
+++ b/leiningen-core/pom.xml
@@ -14,7 +14,7 @@
     </license>
   </licenses>
   <scm>
-    <tag>f42a5cfec3c3bbe1fcc7cb63130955519bb7ca3a
+    <tag>bbbead38b2bbf6e4e7bb315b67dbde6948a6eb84
 </tag>
     <url/>
   </scm>
@@ -106,7 +106,7 @@
     <dependency>
       <groupId>org.apache.maven.wagon</groupId>
       <artifactId>wagon-http</artifactId>
-      <version>2.10</version>
+      <version>2.12</version>
     </dependency>
     <dependency>
       <groupId>com.hypirion</groupId>
@@ -117,6 +117,11 @@
       <groupId>pedantic</groupId>
       <artifactId>pedantic</artifactId>
       <version>0.2.0</version>
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-nop</artifactId>
+      <version>1.7.22</version>
     </dependency>
     <dependency>
       <groupId>org.clojure</groupId>
--- a/leiningen-core/src/leiningen/core/project.clj
+++ b/leiningen-core/src/leiningen/core/project.clj
@@ -49,13 +49,13 @@
   [profile]
   (vector? profile))
 
-(defn artifact-namespace
+(defn group-id
   [id]
   (if (string? id)
     (first (str/split id #"/"))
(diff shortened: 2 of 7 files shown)


leiningen-core/pom.xml
```diff
@@ -106,7 +106,7 @@
     <dependency>
       <groupId>org.apache.maven.wagon</groupId>
       <artifactId>wagon-http</artifactId>
-      <version>2.10</version>
+      <version>2.12</version>
```
danielcompton: Does this need to be bumped too?

cap10morgan: I have no idea what did that. Happy to revert `pom.xml` changes if requested.

leiningen-core/src/leiningen/core/project.clj
```diff
@@ -48,10 +49,22 @@
   [profile]
   (vector? profile))
 
+(defn artifact-namespace
```
danielcompton: This should probably be called `group-id`

leiningen-core/src/leiningen/core/project.clj
```diff
@@ -48,10 +49,22 @@
...
+  [id]
+  (if (string? id)
+    (first (str/split id #"/"))
+    (or (namespace id) (name id))))
+
+(defn artifact-name
```
danielcompton: And artifact-id?

leiningen-core/test/leiningen/core/test/project.clj
```diff
@@ -36,11 +36,13 @@
                :eval-in :leiningen,
                :license {:name "Eclipse Public License"}
 
-               :dependencies '[[leiningen-core/leiningen-core "2.0.0-SNAPSHOT"]
+               :dependencies `[[leiningen-core/leiningen-core "2.0.0-SNAPSHOT"]
```
danielcompton: Do we need to switch the backticks?

leiningen-core/pom.xml
```diff
@@ -119,6 +119,11 @@
       <version>0.2.0</version>
     </dependency>
     <dependency>
+      <groupId>org.slf4j</groupId>
```
danielcompton: Do we need to add this?

cap10morgan: I also have no idea what added this. I didn't intentionally edit the `pom.xml`.

leiningen-core/test/leiningen/core/test/project.clj
cap10morgan: I thought it was needed for the subsequent `~` I added. But maybe that's not right?

cap10morgan:
@danielcompton Good feedback, thanks. I could use some advice on the mysterious `pom.xml` changes. But otherwise have I addressed your concerns?

leiningen-core/test/leiningen/core/test/project.clj
danielcompton: I think you're right, was just checking.

danielcompton:
LGTM

hyPiRion:
Thanks! Don't worry about the pom.xml stuff – We try to keep the pom.xml inside the repo to ease work for upstream maintainers (brew/apt et al), but it also has the side effect that changes are sometimes unintentionally updated. If you leave out any pom.xml changes out next time then that would be great :)

PR #2243, opened 2017-02-02
Add note about --no-sign option in vcs tag docstring

nberger:
This way the option can be found with `lein help vcs tag`


hyPiRion:
Thanks!

PR #2247, opened 2017-02-26
Introduce optional chronological versioning to Leiningen

yegortimoshenko:
Inspired by [Spec-ulation keynote](https://youtu.be/oyLBGkS5ICk?t=47m31s). When version argument is missing in defproject, uses `yyyyMMdd.HHmmss` timestamp in UTC timezone as the project version.

### DIFF
--- a/leiningen-core/src/leiningen/core/project.clj
+++ b/leiningen-core/src/leiningen/core/project.clj
@@ -12,7 +12,9 @@
             [leiningen.core.classpath :as classpath]
             [clojure.string :as str])
   (:import (clojure.lang DynamicClassLoader)
-           (java.io PushbackReader Reader)))
+           (java.io PushbackReader Reader)
+           (java.text SimpleDateFormat)
+           (java.util Date TimeZone)))
 
 (defn make-project-properties [project]
   (with-open [baos (java.io.ByteArrayOutputStream.)]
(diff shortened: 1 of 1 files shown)


hyPiRion:
Thanks for the PR :)

I like the idea of pushing either chronological versioning or just an incremental one (i.e. we're on v44, next will be v45 no matter what), but I'm not sure omitting the version would be the right answer. If you need to deploy to two locations, you will push different versions, although they are semantically the same (and you want them to be the same as well!).

Perhaps it would be better to support this in the `release` task instead? Say we detect that the version string is on the format `YYYYMMDD.HHmm[-suffix]`, then we can change the version before the release. I'm not sure how we would handle snapshots then though, and that's important to support without surprising results.

Happy to hear options and alternatives on this, just bear in mind it shouldn't affect users using semver.

technomancy:
Cool; I agree that the release task is a better place for this than defproject.

danielcompton:
> I'm not sure how we would handle snapshots then though, and that's important to support without surprising results.

Could this work the same as SNAPSHOT support in lein release works currently? i.e. `YYYYMMDD.HHmm-SNAPSHOT` would be the projects version. If the user runs `lein deploy` or `lein install`, the current timestamp is embedded and `20160227.0825-SNAPSHOT` would be deployed. If the user runs `lein release`, then the timestamp is embedded and `20160305.0916` would be deployed.

I'm not really sure how you could resolve the conflict between SNAPSHOT's to reference a stable version identifier which can be continually updated with new SNAPSHOT versions (current behaviour/usage), and the desire for release versions to have a dynamic version which is only assigned on deploy.

technomancy:
I'm going to close this out as the current implementation doesn't seem to be meaningfully distinct to just setting the version to `"SNAPSHOT"`, but of course the door is open to improving support for this inside the `release` task instead.

PR #2250, opened 2017-03-01
Make profiles string in target path smarter

weavejester:
Currently if Leiningen is passed a target path like:

```clojure
:target-path "target/%s/"
```

Then the `%s` is substituted with a string of profiles. For example:

```
target/base+system+user+provided+dev
```

However, because the `:default` profile is a collection of all of these, it would be more concise for the target path to be:

```
target/default
```

This PR alters the `profile-scope-target-path` function to replace subsets of profiles that match a composite profile. So for example:

```clojure
[:base :system :user :provided :dev]      => "default"
[:base :system :user :provided :dev :foo] => "default+foo"
[:base :system :user :provided]           => "base+system+user+provided"
```

Currently composite profiles are added to the beginning of the string. Should order matter in this case?

### DIFF
--- a/leiningen-core/src/leiningen/core/project.clj
+++ b/leiningen-core/src/leiningen/core/project.clj
@@ -467,11 +467,24 @@
   (.toString (BigInteger. 1 (-> (java.security.MessageDigest/getInstance "SHA1")
                                 (.digest (.getBytes content)))) 16))
 
+(defn- keyword-composite-profile? [profile]
+  (and (composite-profile? profile) (every? keyword? profile)))
+
+(defn- normalize-profile-names [project profiles]
+  (reduce
+   (fn [profiles [composite keys]]
+     (if (set/subset? (set keys) (set profiles))
+       (->> profiles (remove (set keys)) (cons composite))
+       profiles))
+   profiles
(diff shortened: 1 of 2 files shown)


hyPiRion:
Theoretically, the order would matter; `[:base :system :user :provided :dev :foo]` would be a different profile than, say, `[:foo :base :system :user :provided :dev]` and `[:base :system :foo :user :provided :dev]`, as the merge logic depends on ordering.

I wouldn't mind a version which detects composite profiles and merges them together, as long as they're put inside the correct location.

weavejester:
I thought that might be the case, now that I've had time to think about it as well. Rather than looking for subsets, I'll look for subsequences and update the PR accordingly.

weavejester:
What about that? It performs a greedy match to find ordered subsequences that match a composite. The profiles are iterated in order, and the longest composites are tried first.

weavejester:
Actually, hold off on merging this. It probably wants to resolve profiles recursively.

weavejester:
Okay, now composite profiles that depend on other composite profiles are recursively normalized when producing the target path.

technomancy:
Looks good; thanks.

PR #2254, opened 2017-03-06
Removed duplicate of alias clojure.string in leiningen-core's project.clj

Rovanion:
It was aliased both as `s` and `str`, I removed `s` and kept `str`.


hyPiRion:
Thanks :)

Rovanion:
Cheers!

PR #2255, opened 2017-03-07
Provide tree deps in data format for display/parsing by other tools

kenrestivo-stem:
Per @technomancy 's suggestion in IRC.


technomancy:
Once you fix the tests if you could get rid of the spurious whitespace changes that would make this patch a bit easier to review; thanks!

PR #2256, opened 2017-03-11
Provide tree deps in EDN format for display/parsing by other tools

kenrestivo-stem:
Support outputting tree data as EDN for parsing by other programs.

### DIFF
--- a/src/leiningen/deps.clj
+++ b/src/leiningen/deps.clj
@@ -4,6 +4,7 @@
             [leiningen.core.main :as main]
             [leiningen.core.project :as project]
             [leiningen.core.user :as user]
+            [clojure.pprint :as pprint]
             [leiningen.core.utils :as utils]
             [cemerick.pomegranate.aether :as aether])
   (:import (org.sonatype.aether.resolution DependencyResolutionException)))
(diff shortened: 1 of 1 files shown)


src/leiningen/deps.clj
```diff
@@ -123,7 +130,10 @@ force them to be updated, use `lein -U $TASK`."
                               dependencies-key
                               managed-dependencies-key
                               project)]
-               (walk-deps hierarchy print-dep))
+               (case command
+                 ":tree" (walk-deps hierarchy print-dep)
+                 ":tree-data"  (binding [*print-length* 10000 *print-level* 10000]
```
danielcompton: Looking at the failed tests, it seems like you also need to handle `:plugin-tree` here?

danielcompton:
It looks like this is the successor to #2255.

src/leiningen/deps.clj
danielcompton: Also, that means `:plugin-tree-data` should probably also be provided?

kenrestivo-stem: OK, will do.

kenrestivo-stem:
Should be ready to go, thanks.

technomancy:
Looks great; thanks!

PR #2257, opened 2017-03-11
give reference to how to get a SNAPSHOT version

timofeytt:


doc/TUTORIAL.md
```diff
@@ -261,7 +261,8 @@ wider JVM community.
 You can add third-party repositories by setting the `:repositories` key
 in project.clj. See the
 [sample.project.clj](https://github.com/technomancy/leiningen/blob/stable/sample.project.clj)
-for examples on how to do so.
+for examples on how to do so. For example, you may want to add the Sonatype repository in 
+order to get the latest SNAPSHOT development version of a library.
```
danielcompton: If this sentence is to be added, you probably need to give context on why someone would need to add the Sonatype repository to get SNAPSHOTs, which repository that is, and which SNAPSHOT JARs would be in the sonatype snapshot repo. For example Clojars stores SNAPSHOTs together with releases, so it's not needed for this.

Perhaps something more like:

> For example, if you want a SNAPSHOT version of a Java library you may need to add the Sonatype snapshots repository (https://oss.sonatype.org/content/repositories/snapshots/), as snapshots and releases are deployed to separate repositories.

Overall, I'm not convinced that this sentence makes sense here. If someone is wanting to use a SNAPSHOT Java library from Clojure then I would assume that they understand about the different Maven repositories available. Also, any library that is offering a SNAPSHOT version is probably going to put in their README which repo SNAPSHOTs are deployed to.

timofeytt: I want to get a SNAPSHOT version of clojure.data.xml. I don't know much about repos. I've looked on this page and found only Maven Central and Clojars both of which lack the desired version. I had to ask on Stackoverflow to find out that Sonatype holds snapshots of Clojure libs! I think it makes sense here, because there is a link to the sample.project which uses Sonatype.

danielcompton: Gotcha, forgot about Clojure snapshots. If you add that context, I think this could make a good addition to the docs.

timofeytt:
Is this ok with you?

danielcompton:
Looks ok to me, but I'm not a committer, will let them chip in.

technomancy:
Looks great; thanks.

PR #2258, opened 2017-03-12
Catch NoRouteToHostException exception

### LINKED ISSUE #2054: lein commands should work when a source is down
clojars.org is down right now. As a result, running lein stalls (seemingly forever) with this output:

$ lein repl
Jan 01, 2016 12:57:57 PM org.apache.http.impl.execchain.RetryExec execute
INFO: I/O exception (java.net.NoRouteToHostException) caught when processing request to {s}->https://clojars.org:443: No route to host
Jan 01, 2016 12:57:57 PM org.apache.http.impl.execchain.RetryExec execute
INFO: Retrying request to {s}->https://clojars.org:443
Jan 01, 2016 12:59:12 PM org.apache.http.impl.execchain.RetryExec execute
INFO: I/O exception (java.net.NoRouteToHostException) caught when processing request to {s}->https://clojars.org:443: No route to host
Jan 01, 2016 12:59:12 PM org.apache.http.impl.execchain.RetryExec execute
INFO: Retrying request to {s}->https://clojars.org:443
Jan 01, 2016 1:00:28 PM org.apache.http.impl.execchain.RetryExec execute
INFO: I/O exception (java.net.NoRouteToHostException) caught when processing request to {s}->https://clojars.org:443: No route to host
...

A workaround is to use LEIN_OFFLINE=1:

```
LEIN_OFFLINE=1 lein ...
```

This is poor developer usability. This shouldn't be necessary if all dependencies are fetched already, and it's harder / less obvious how to set this env variable from within editors like Vim and Emacs for Fireplace and Cider workflows.

firesofmay:
In case network is down while fetching dependencies and we get
NoRouteToHostException exception, fallback to `:offline? true`
behaviour.

Fixes #2054 issue.

### DIFF
--- a/leiningen-core/src/leiningen/core/classpath.clj
+++ b/leiningen-core/src/leiningen/core/classpath.clj
@@ -298,10 +298,12 @@
          (throw (ex-info "Could not resolve dependencies" {:suppress-msg true
                                                            :exit-code 1} e)))
        (catch Exception e
-         (if (and (instance? java.net.UnknownHostException (root-cause e))
-                  (not offline?))
-           (get-dependencies-memoized dependencies-key (assoc project :offline? true))
-           (throw e)))))))
+         (let [exception-cause (root-cause e)]
+           (if (and (or (instance? java.net.UnknownHostException exception-cause)
+                        (instance? java.net.NoRouteToHostException exception-cause))
+                    (not offline?))
+             (get-dependencies-memoized dependencies-key (assoc project :offline? true))
+             (throw e))))))))
 
 (defn- group-artifact [artifact]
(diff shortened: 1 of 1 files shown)


firesofmay:
@danielcompton Pushed a new patch as suggested.

firesofmay:
Any updates on this? Is this good to merge?

danielcompton:
I'm not a committer, so I can't merge this. If you wanted to be more precise, you could catch `java.io.IOException` which is a superclass of NoRouteToHostException and UnknownHostException, but I'll let the maintainers make a judgement call on that.

technomancy:
Looks good to me. Thanks!

PR #2261, opened 2017-03-28
Make repls outside project run in Leiningen

### LINKED ISSUE #1900: lein not using nrepl 0.2.7 even when latter is installed
`lein version` reports
```
Leiningen 2.5.1 on Java 1.7.0_79 OpenJDK 64-Bit Server VM
```
My `~/.lein/profiles.clj` looks like:
```clj
{:user {:dependencies ^:replace [[org.clojure/tools.nrepl "0.2.7"]]
        :plugins [[cider/cider-nrepl "0.9.0-SNAPSHOT"]
                  [lein-pprint "1.1.1"]]}}
```
and indeed, `~/.m2/repository/org/clojure/tools.nrepl/` contains `0.2.7/tools.nrepl-0.2.7.jar` and `0.2.7/tools.nrepl-0.2.7.jar.sha1`. However, running `lein repl` from `~/` still produces:
```
nREPL server started on port 57962 on host 127.0.0.1 - nrepl://127.0.0.1:57962
REPL-y 0.3.5, nREPL 0.2.6
Clojure 1.6.0
```
i've come across this issue whilst trying to upgrade nREPL for CIDER 0.9.0snapshot. In [this SO thread about trying to upgrade to nrepl-0.2.7](http://stackoverflow.com/q/28688721), one user [says](http://stackoverflow.com/a/30139330) that the dependency is only enabled by `lein` if one is running `lein repl` from within a Clojure project. Is this the intended behaviour? Or might there be something wrong with the rest of my setup, such that the `:dependencies` vector in `~/.lein/profiles.clj` is effectively ignored?

technomancy:
In order to fix #1900, a while back we made it so that out-of-project repls *always* run in their own subprocess, which increases load time and breaks tools like Grenchman.

This PR rolls back that change and addresses the real root problem; instead of forcefully applying the workaround it warns the user what happened and suggests the workaround.


technomancy:
@hyPiRion interested in your feedback on this.

winks:
I suppose you painstakingly created this `bootclasspath-deps.clj` by manually checking all the deps. I also suppose there's no decent way to generate that file and/or extract the specific versions from `project.clj` or something on buildtime.

technomancy:
Yeah, my least-favorite part of this is definitely the bit where we have to keep two lists of dependencies in sync. We can write a small tool that doesn't get distributed with lein itself (in bin/ or something?) to automate it, but remembering to run it...

On the other hand, it's not the end of the world if the list gets a bit outdated.

PR #2262, opened 2017-03-30
Found a redundency in code and reduced it

Rovanion:
While reading the code I found the same snippet of code being used here as the functions in `leiningen.core.project` so I simply made the core in `leiningen.pom` use that.


technomancy:
Looks good; thanks!

PR #2263, opened 2017-04-04
add linux commands for installation to readme

light24bulbs:
found these on another open source project.  They make it much faster to install on ubuntu, especially if you're a little rusty with bash.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,39 @@ separate install script.  Follow these instructions to install Leiningen manuall
 4. Set it to be executable. (`chmod 755 ~/bin/lein`)
 5. Run it.
 
+### Ubuntu commands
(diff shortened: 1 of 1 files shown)


technomancy:
There are a few problems with this--I don't want to recommend a random PPA over distro-specific packages, and I strongly recommend against using any Oracle software ever. Including instructions that only work on Ubuntu doesn't really make sense; next someone will want to add Fedora, then Debian, Arch, Gentoo--we can't cover them all, so generic advice like "make sure Java is installed" is better here. Perhaps we could link to a wiki page if people need more details, but the readme is already too long as it is.

light24bulbs:
Makes sense! Thanks for the explanation!

PR #2266, opened 2017-04-18
Extract vcs-dir from leiningen.vcs/uses-vcs

yegortimoshenko:
I'd like to use `vcs-dir` in my Leiningen plugin.


technomancy:
Sorry for the delay here; I don't mind merging this patch if the tests pass.

PR #2267, opened 2017-04-18
Make javac respect LEIN_SILENT.

cespare:


technomancy:
Good catch; thank you!

Issue #2272, opened 2017-05-09
lein repl fails if profiles.clj file is empty


iperdomo:
If you have an empty file in `$HOME/.lein/profiles.clj`, `lein repl` fails:

````
$ echo -n '' > ~/.lein/profiles.clj

$ lein repl

Error reading profiles.clj from /home/ivan/.lein
java.lang.RuntimeException: EOF while reading
 at clojure.lang.Util.runtimeException (Util.java:221)
    clojure.lang.LispReader.read (LispReader.java:246)
    clojure.lang.LispReader.read (LispReader.java:196)
    clojure.lang.LispReader.read (LispReader.java:185)
    clojure.lang.RT.readString (RT.java:1835)
    clojure.lang.RT.readString (RT.java:1830)
    clojure.core$read_string.invokeStatic (core.clj:3687)
    clojure.core$read_string.invoke (core.clj:3677)
    leiningen.core.utils$read_file.invokeStatic (utils.clj:43)
    leiningen.core.utils$read_file.invoke (utils.clj:39)
    leiningen.core.user$fn__6513.invokeStatic (user.clj:69)
    leiningen.core.user/fn (user.clj:68)
    clojure.lang.AFn.applyToHelper (AFn.java:154)
    clojure.lang.AFn.applyTo (AFn.java:144)
    clojure.core$apply.invokeStatic (core.clj:646)
    clojure.core$memoize$fn__5708.doInvoke (core.clj:6107)
    clojure.lang.RestFn.invoke (RestFn.java:408)
    leiningen.core.user$fn__6516.invokeStatic (user.clj:87)
    leiningen.core.user/fn (user.clj:77)
    clojure.lang.AFn.applyToHelper (AFn.java:152)
    clojure.lang.AFn.applyTo (AFn.java:144)
    clojure.core$apply.invokeStatic (core.clj:646)
    clojure.core$memoize$fn__5708.doInvoke (core.clj:6107)
    clojure.lang.RestFn.invoke (RestFn.java:397)
    leiningen.core.user$profile_auth.invokeStatic (user.clj:193)
    leiningen.core.user$profile_auth.invoke (user.clj:190)
    leiningen.core.classpath$add_repo_auth.invokeStatic (classpath.clj:210)
    leiningen.core.classpath$add_repo_auth.invoke (classpath.clj:194)
    clojure.core$map$fn__4785.invoke (core.clj:2644)
    clojure.lang.LazySeq.sval (LazySeq.java:40)
[...]

PR #2273, opened 2017-05-09
[#2272] Catch exceptions on malformed profiles.clj files

iperdomo:
Using clojure.edn/read-string results in nil instead of an exception

    user=> (edn/read-string "")
    nil
    user=> (read-string "")
    RuntimeException EOF while reading  clojure.lang.Util.runtimeException (Util.java:221)

### DIFF
--- a/leiningen-core/src/leiningen/core/utils.clj
+++ b/leiningen-core/src/leiningen/core/utils.clj
@@ -1,6 +1,7 @@
 (ns leiningen.core.utils
   (:require [clojure.java.io :as io]
-            [clojure.java.shell :as sh])
+            [clojure.java.shell :as sh]
+            [clojure.edn :as edn])
   (:import (com.hypirion.io RevivableInputStream)
            (clojure.lang LineNumberingPushbackReader)
            (java.io ByteArrayOutputStream PrintStream File FileDescriptor
@@ -40,7 +41,7 @@
   "Returns the first Clojure form in a file if it exists."
   [file]
   (if (.exists file)
-    (try (read-string (slurp file))
+    (try (edn/read-string (slurp file))
         (catch Exception e
          (binding [*out* *err*] ;; TODO: use main/warn for this in 3.0
            (println "Error reading"
--- a/leiningen-core/src/leiningen/core/utils.clj
+++ b/leiningen-core/src/leiningen/core/utils.clj
@@ -1,7 +1,6 @@
 (ns leiningen.core.utils
   (:require [clojure.java.io :as io]
-            [clojure.java.shell :as sh]
-            [clojure.edn :as edn])
+            [clojure.java.shell :as sh])
   (:import (com.hypirion.io RevivableInputStream)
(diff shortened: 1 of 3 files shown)


iperdomo:
Not sure why the tests failed. `bin/lein test` works on my local environment ... The full log:

https://gist.github.com/iperdomo/c71ae88e7247e070b331e7b78df9d869

technomancy:
Sorry; I'm afraid we can't change this--clojure.edn/read-string does not
work properly when the file contains read-eval forms. Changing it right
now (even if it's the right thing to do to remove this feature) would be
the wrong time for a breaking change.

iperdomo:
@technomancy yes, i understand that both functions don't have the same semantic.

Would you accept catching the EOF exception instead? It was very unfortunate first issue found by a new Clojure user following a tutorial. Instead of removing the file he removed the content and saved it. After that, the `lein repl` was not working. An experienced Clojure/Java developer understands the stack trace, but it was hard for him to understand why this was failing.

technomancy:
Yes; sorry, I should have been clearer about that. Catching the EOF exception and emitting a sensible error message instead of an enormous stack trace would be great.

iperdomo:
Empty file

    $ echo -n '' > ~/.lein/profiles.clj 
    $ ./bin/lein repl
    Error reading profiles.clj from /home/ivan/.lein
    File cannot be empty
    nREPL server started on port 32793 on host 127.0.0.1 - nrepl://127.0.0.1:32793
    REPL-y 0.3.7, nREPL 0.2.12
    [...]

Invalid content - 1

    $ echo -n '{:user' > ~/.lein/profiles.clj 
    $ ./bin/lein repl
    Error reading profiles.clj from /home/ivan/.lein
    Invalid content was found
    nREPL server started on port 38129 on host 127.0.0.1 - nrepl://127.0.0.1:38129
    REPL-y 0.3.7, nREPL 0.2.12
    [...]

Invalid content - 2

    $ echo -n '{:user}' > ~/.lein/profiles.clj
    $ ./bin/lein repl
    Error reading profiles.clj from /home/ivan/.lein
    Map literal must contain an even number of forms
    nREPL server started on port 33941 on host 127.0.0.1 - nrepl://127.0.0.1:33941
    REPL-y 0.3.7, nREPL 0.2.12

iperdomo:
@technomancy not sure why those test are failing :disappointed: .... they work on one of builds in travis-ci though (https://travis-ci.org/technomancy/leiningen/jobs/231054052)

If you take this changeset, i guess you can squash when merging, so discard the previous attempt.

hyPiRion:
@iperdomo: This has nothing to do with your commit afaict. The issue seems to be cause by the protocol version of TLS from clojars, although I'm not sure why this is the case.

iperdomo:
@hyPiRion ok, good to know

technomancy:
Yeah, these failures are definitely not related to your changes. Thanks!

PR #2285, opened 2017-06-14
Fix #2284: avoid visiting checkouts we've already seen

### LINKED ISSUE #2284: Recursive checkouts are slow and inefficient
The function `checkout-deps-paths` is recursively called because it's included in `:checkout-deps-shares` for the default project profiles. This allows leiningen to recursively build a checkouts classpath by traversing all projects reachable from checkouts, and their checkouts (etc).

In practice, it's pretty reasonably to assume that checkouts will be a directed-acyclic-graph, rather than a tree, so this form of recursion results in the DAG being fully expanded into a tree, creating a lot of duplication. In our project we have 11 projects with a total of 37 checkouts. Calling `lein cp` on the root project takes 13 seconds and creates over 1000 classpath entries for 75 unique directories. This is not only unacceptably slow for starting leiningen, but results in an unwieldy classpath that causes poor class resolution performance for Java.

I think `checkout-deps-paths` needs to be graph-aware – it should use a recursion mechanism that eliminates projects that it has already seen, and it should use `getCanonicalFile` to remove symlinks from classpath entries.

sfnelson:
Checkouts are recursive by default and the same checkout project may be
reachable via several paths when symlinks are used. Avoid visiting projects
multiple times by keeping track of seen projects while accumulating
checkouts classpath entries.

### DIFF
--- a/leiningen-core/src/leiningen/core/classpath.clj
+++ b/leiningen-core/src/leiningen/core/classpath.clj
@@ -541,20 +541,38 @@
         (resolve-managed-dependencies :dependencies :managed-dependencies)
         (map (memfn getAbsolutePath)))))
 
+(defn- visit-project!
+  "Records a visit to a project into the volatile `seen`, returning nil if the project has already been visited."
+  [seen {:keys [root] :as project}]
+  (when-not (@seen root)
+    (vswap! seen conj root)
+    project))
+
+(defn- project-paths
+  "Returns a function that applies each function in checkout-paths to a given project and returns a flattened list of
+  classpath entries."
+  [checkout-paths]
+  (if (seq checkout-paths)
+    (comp flatten (apply juxt checkout-paths))
+    (constantly nil)))
+
+(def ^:internal ^:dynamic *seen* nil)
+
 (defn ^:internal checkout-deps-paths
   "Checkout dependencies are used to place source for a dependency
   project directly on the classpath rather than having to install the
   dependency and restart the dependent project."
-  [project]
+  [{:keys [checkout-deps-shares root] :as project}]
   (require 'leiningen.core.project)
   (try
-    (let [checkout-paths (:checkout-deps-shares project)
-          checkouts ((resolve 'leiningen.core.project/read-checkouts) project)]
-      (mapcat (fn [checkout]
-                ;; can't mapcat here since :checkout-deps-shares points to
-                ;; vectors and strings
-                (flatten (map #(% checkout) checkout-paths)))
-              checkouts))
+    ;; This function needs to be re-entrant as it is one of the default members of `:checkout-deps-shares`.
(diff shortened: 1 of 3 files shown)


sfnelson:
This isn't working properly yet, I don't have more time to work on it now so I'll let you know when it's ready.

sfnelson:
@technomancy ready for review

sfnelson:
With these changes my project takes 3 seconds to compute `lein cp`

technomancy:
Nice work; thanks! One thing I noticed is that `*seen*` doesn't really need to be dynamic in the current implementation; do you think you could replace it with a local?

While you're at it, maybe you could replace the internal `fn` with a named `defn-` which is partially applied to the `seen` volatile that would tidy it up a bit; I feel like the `checkout-deps-paths` body is getting a bit long.

Thanks.

technomancy:
@hyPiRion hm; I thought we dropped support for 6 a while back.

sfnelson:
@technomancy

>  One thing I noticed is that *seen* doesn't really need to be dynamic in the current implementation; do you think you could replace it with a local?

The recursion mechanism uses the `checkout-deps-path` var directly so there's no way for me to share state between invocations without using a var myself (unless I'm missing something)? If it weren't for this I'd favour passing around an accumulator so that the traversal doesn't need to be eager and we don't need mutable state. How do you feel about taking the recursion var out of `:checkout-deps-shares`? Does anyone actually want to change `:checkout-deps-shares` and *doesn't* want recursion?

@hyPiRion I'm happy to play import games to make the tests Java 6 compatible, I wasn't sure if that was a goal anymore. The README states that Java 6 is supported but the automated tests don't run on Java 6.

technomancy:
Oh, I see; I was thinking it would be enough to make it an additional arg to `checkout-deps-shares`, but that would miss some cases where a different branch of a tree re-used a checkout.

Probably the inner `fn` could still become a top-level `defn` that gets partially applied tho; what do you think about that?

sfnelson:
@technomancy @hyPiRion I've made the changes you suggested

technomancy:
Looks great; thanks.

hyPiRion:
The latest lein release was built with JDK 1.6, so Lein is still 1.6-compatible for now. I wouldn't mind moving up to 1.7, as there are a lot of new useful `nio` functions and some "new" useful pipe redirection mechanisms.

A few percent of Clojure programmers* uses 1.6 or lower, and Clojure still works on 1.6, so if we decide to ditch 1.6, we should at least put it in the changelog.

*According to [State of Clojure 2016](http://blog.cognitect.com/blog/2017/1/31/state-of-clojure-2016-results), it's less than 5% of people though.

technomancy:
I must have been misremembering. We can revisit the question when we come across a situation where graceful fallback isn't so simple.

PR #2289, opened 2017-07-13
Trim whitespace in Git HEAD commit hash

glts:
The Git commit hash in both the generated pom.xml and the pom.properties file includes a trailing newline, but only when the code path is taken that runs `git rev-parse`. Other code paths (`read-git-ref`, `read-git-head-file`) already trim what they read so do it here as well.

Have to make the same change twice because of duplicated code.

No tests, but verified locally. Thank you!


hyPiRion:
Thanks for this fix :) The patch looks sound, and I can't imagine this breaking anything.

glts:
Thank you, @hyPiRion!

PR #2291, opened 2017-07-21
Do not include project.clj in JAR root (reverts commit 0412001)

glts:
Consider reverting commit 0412001c, made five years ago for Leiningen 1 compatibility.

In commit d038306 (2012-04-12) JAR format was tidied up a little to contain `project.clj` in a namespaced dir in `META-INF/leiningen` instead of directly in the root (and therefore on the classpath).

In commit 0412001 (2012-06-22), before the release of Leiningen 2, this change was partially undone to retain compatibility with Leiningen 1.

Is Leiningen 1 still fully supported? If not, please consider reverting 0412001 to avoid shadowing issues/not have project.clj included multiple times/not have project.clj on the classpath/slightly tidier JARs.

### DIFF
--- a/src/leiningen/jar.clj
+++ b/src/leiningen/jar.clj
@@ -199,8 +199,6 @@
                             (:group project) (:name project))
               :bytes (.getBytes (pom/make-pom project))}
              {:type :bytes :path (scope "project.clj")
-              :bytes (.getBytes (slurp (str (:root project) "/project.clj")))}
-             {:type :bytes :path "project.clj"
               :bytes (.getBytes (slurp (str (:root project) "/project.clj")))}]
             (for [doc (map (partial io/file (:root project))
                         (concat readmes licenses))


hyPiRion:
I think it's safe to consider Lein 1 as not supported – there might be people which still rely on this, but that's Hyrum's law in action. As we do have good documentation on how to insert `project.clj` into jars in the FAQ, I don't see any reason to not remove this functionality unless we have somehow managed to accidentally depend on it.

glts:
Agree, this is mainly about default behaviour. As it is, `project.clj` is eventually included on the classpath. Doing `(require 'project)` when a Leiningen-built JAR is around gives

    CompilerException java.lang.RuntimeException: Unable to resolve symbol: defproject in this context, compiling:(project.clj:1:1)

instead of

    FileNotFoundException Could not locate project__init

as for any other unknown namespace.

This may not be a real issue but it is also not useful or ‘correct’ behaviour. Hence the suggestion to drop `project.clj` in root and go with the original change d038306. Cheers!

hyPiRion:
Yeah, I agree. It's more about what users expect in uberjars. It's very easy to do

```clj
(require '[clojure.java.io :as io])
(println (slurp (io/resource "project.clj")))
```

And, as if by magic, you have the project.clj.

Since it's also available in `META-INF/leiningen/group-id/artifact-id/project.clj` I think I'll just update the FAQ to include the location of project.clj for those interested, then merge this in.

Issue #2236, opened 2017-01-21
leiningen.core.user/gpg map returned is inconsistent


miikka:
```
% lein deps :verify
gpg: assuming signed data in '/Users/miikka/.m2/repository/instaparse/instaparse/1.4.0/instaparse-1.4.0.jar'
gpg: Signature made Ke  6 Tou 11:48:18 2015 EEST
gpg:                using RSA key 0x1B5C6189B92AC759
gpg: Can't check signature: No public key
java.lang.NullPointerException: null
 at java.util.regex.Matcher.getTextLength (Matcher.java:1283)
    java.util.regex.Matcher.reset (Matcher.java:309)
    java.util.regex.Matcher.<init> (Matcher.java:229)
    java.util.regex.Pattern.matcher (Pattern.java:1093)
    clojure.core$re_matcher.invokeStatic (core.clj:4674)
    clojure.core$re_find.invokeStatic (core.clj:4716)
    clojure.core$re_find.invoke (core.clj:4716)
    leiningen.deps$fetch_key.invokeStatic (deps.clj:28)
    leiningen.deps$fetch_key.invoke (deps.clj:27)
    leiningen.deps$check_signature.invokeStatic (deps.clj:41)
    leiningen.deps$check_signature.invoke (deps.clj:37)
 (rest of the stacktrace omitted)
```

In this example it happens because the public key is not found, but [looking at the code](https://github.com/technomancy/leiningen/blob/a76de08f0da4c5ca95ceaf1421f0122b2df044a1/src/leiningen/deps.clj#L37-L41), it happens if `gpg` returns a non-zero exit code for any reason while checking a signature. The trouble is that [`user/gpg` does not return stderr in `:err`](https://github.com/technomancy/leiningen/blob/master/leiningen-core/src/leiningen/core/user.clj#L106-L122) anymore. This means that even if `fetch-key` could handle a nil, it wouldn't fetch any GPG keys because the error message detection wouldn't work.

LoganGirard:
I've been looking into this. I'm able to repro this on stable but not on master. Can anyone else corroborate that? Seems like this was somehow fixed in a checkin, but I haven't been able to identify what check in that is or why.

miikka:
Hey, thanks for looking into this! I just tried out this on master and I still get the error. If you're reproducing this in the Leiningen repo, I wonder if it's possible that the dependencies have changed and you have all the public keys for them, so that this error does not happen anymore.

[...]

PR #2293, opened 2017-07-29
Return consistent result from leiningen.core.user/gpg

glts:
Change `user/gpg` to always return maps containing the entries `:out`
and `:err` with the contents of the corresponding stdout/stderr streams.
stderr is also copied to `System/err` as before.

Proposed fix for #2236. Verified locally with gpg/gpg2.

Thank you!


hyPiRion:
This looks reasonable to me, at least :)

technomancy:
Cool; build failure seems to be a network problem with Maven Central that is unrelated to your change. Thanks!

glts:
Cool, thank you!

PR #2294, opened 2017-08-04
Correct typo in README.md

cbcoutinho:
Upgrading lein to the latest stable version is done using `upgrade`, not `update` - this should be reflected in the README.


hyPiRion:
Nice catch, thanks!

PR #2297, opened 2017-08-10
Update manpage

ehashman:
As I need a manpage to package the leiningen application for Debian, I was pleased to find that one already existed... but then I noticed a few typos and wanted to freshen it up a little :)


hyPiRion:
Thank you, this looks very good :)

Issue #2278, opened 2017-06-01
New version of Aether/pomegranate breaks s3-wagon-private


technomancy:

technomancy:
The weird thing about this is that the tests for pomegranate show that registering a custom wagon works fine under some circumstances: https://github.com/cemerick/pomegranate/blob/master/src/test/clojure/cemerick/pomegranate/aether_test.clj#L591

For some reason the S3 wagon is different. We register it here: https://github.com/cemerick/pomegranate/blob/master/src/main/clojure/cemerick/pomegranate/aether.clj#L114

The old implementation which works with the S3 wagon is here: https://github.com/cemerick/pomegranate/blob/pomegranate-0.3.1/src/main/clojure/cemerick/pomegranate/aether.clj#L109

ghadishayban:
I couldn't reproduce this on a bare project that used lein 2.7.2-SNAPSHOT compiled this morning.  I was able to successfully push and pull a dep from a private s3 repo

technomancy:
@ghadishayban could you let me know what version of s3-wagon-private you used? Seems I get different results with 1.2.0 and 1.3.0 (though both failures). The wagon is registered correctly, but the auth info (passphrase) isn't propagated correctly, leading to "Secret key cannot be null" errors.

technomancy:
I believe what's happening is that the authentication settings are set (I've confirmed that Pomegranate calls `setAuthentication` the same way it used to) but that the new version of Aether ignores this authentication data. Perhaps it provides another way to set the credentials in a way that they would actually be used, but since there is no documentation published it's difficult to tell; I believe this would require a deep-dive into the source by someone who is very familiar with reading Java code.

glts:
With the help of @technomancy I managed to find the cause of this,
thankfully not buried very deeply.

Commit cemerick/pomegranate@b71ad6f0e3e88003c1ee28458af1f94bd319f6a6
contains the relevant diff:

```diff
 (defn- set-authentication
   "Calls the setAuthentication method on obj"
   [obj {:keys [username password passphrase private-key-file] :as settings}]
   (if (or username password private-key-file passphrase)
-    (.setAuthentication obj (Authentication. username password private-key-file passphrase))
+    (doto obj
+      (.setAuthentication
+       (.build
+        (doto (AuthenticationBuilder.)
+          (.addUsername username)
+          (.addPassword password)
+          (.addPrivateKey private-key-file passphrase)))))
     obj))
```

[...]

PR #2298, opened 2017-08-12
[WIP] Respect declaration order of dependencies

### LINKED ISSUE #2283: Classpath order is not honored
I need to rely on the fact that a dependency `A` comes before dependency `B` on the classpath. Currently by specifying `A` before `B` in the `:dependencies` vector does not necessarily generate a classpath that honors this. It seems that the dependencies are turned into a `set` before the classpath gets generated. Maven on the other hand does generate a classpath in the same order the dependencies are declared.

glts:
This change modifies Leiningen’s behaviour to respect/preserve the declaration order of dependencies. Dependency order is visible in tasks such as `lein classpath` or `lein deps :tree`.

The Maven dependency resolution mechanism which Leiningen uses handles dependencies in declaration order and preserves declaration order throughout. Leiningen currently does not take this dependency order information into account. Indeed, this information is not available through Pomegranate, the library to which Leiningen delegates the job of resolving dependencies. With this change (and the one proposed for cemerick/pomegranate#86) we fall back in line with the Maven way of doing things.

For users of Leiningen, the current state of affairs is this: dependencies appear sometimes in random order (`lein classpath`), sometimes in a deterministic but arbitrary order (`lein deps :tree` in alphabetic order).

The proposed new state of affairs is this: dependencies appear in the order in which they are declared, consistent with Maven. There is some risk because dependencies are such a fundamental thing – however, I think the change should be transparent since we would be going from an arbitrary order to an obviously deterministic, user-controllable one.

This is ‘work in progress’ because it relies on the proposed (not vetted, not released) change at cemerick/pomegranate#88. 

I have verified locally that this resolves #2283.

### DIFF
--- a/leiningen-core/pom.xml
+++ b/leiningen-core/pom.xml
@@ -90,7 +90,7 @@
     <dependency>
       <groupId>com.cemerick</groupId>
       <artifactId>pomegranate</artifactId>
-      <version>0.4.0-alpha1</version>
+      <version>0.4.0-alpha2-SNAPSHOT</version>
       <exclusions>
         <exclusion>
           <groupId>org.tcrawley</groupId>
--- a/leiningen-core/project.clj
+++ b/leiningen-core/project.clj
@@ -7,7 +7,7 @@
                  [bultitude "0.2.8" :exclusions [org.tcrawley/dynapath]]
                  [org.flatland/classlojure "0.7.1"]
                  [robert/hooke "1.3.0"]
-                 [com.cemerick/pomegranate "0.4.0-alpha1"
+                 [com.cemerick/pomegranate "0.4.0-alpha2-SNAPSHOT"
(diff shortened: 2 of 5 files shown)


glts:
Hey, I don’t mean to push this too hard, just wanted to write it all up while it’s still fresh in my mind.

hyPiRion:
Oh, no worries at all :) I looked through both this and cemerick/pomegranate#86, and it seems pretty straightforward and OK to me – we just need need to get it into a released pomegranate version.

technomancy:
We ought to get a fix for #2278 before the next pomegranate release as well since I don't think that can be fixed in Leiningen alone.

glts:
This is probably not the right solution, see discussion in cemerick/pomegranate#86.

PR #1817, opened 2015-01-29
Add typehint to `class` param passed to Reflector/invokeStaticMethod

ragnard:
Eliminates reflection warnings like:

```
Reflection warning, /tmp/form-init6919407174722938138.clj:1:991 - call to static method invokeStaticMethod on clojure.lang.Reflector can't be resolved
```
when using `lein run` and friends.

PR #2306, opened 2017-09-01
Type-hint leiningen.run/run-form and print metadata in eval file

glts:
#1817 attempted to get rid of a reflection warning in the run task. A
fix was committed in 54fd206600fe204b592e4cfe94d756e6338983ac, but it
didn’t have any effect. Here we enable metadata printing when spitting
the form and add a second type hint.

For repro see my comment at #1817.

This may not be the right fix! I have verified that it works though.


technomancy:
Looks pretty reasonable to me; thanks.

Issue #2308, opened 2017-09-06
Consider allowing templating of the vcs commit task


tobias:
It would be nice if the `vcs commit` task took a template string that allowed overriding the commit message. The current message is just "Version &lt;version&gt;", so the template could be a `format` ready string that takes the version. 

My primary use case for this is to allow using `lein release` on CI without the commits from the release retriggering a CI build, but this would allow other customization use cases as well:

```
:release-tasks [...
                ["vcs" "commit" "Version %s [skip ci]"]
                ...]
```

PR #2312, opened 2017-09-09
Allow overriding commit message

tobias:
This gives the user control over the commit message when overriding the
default :release-tasks.

See #2308

### DIFF
--- a/src/leiningen/vcs.clj
+++ b/src/leiningen/vcs.clj
@@ -28,12 +28,16 @@
       nil parsed-args                                       ;; We're finished and can exit
       (recur (assoc parsed-args :prefix (first args)) (rest args)))))
 
+(def default-commit-message "Version %s")
+
 ;;; Methods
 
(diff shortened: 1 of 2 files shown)


src/leiningen/vcs.clj
```diff
@@ -65,9 +69,10 @@
   (binding [eval/*dir* (:root project)]
     (apply eval/sh-with-exit-code "Couldn't push to the remote" "git" "push" "--follow-tags" args)))
 
-(defmethod commit :git [project]
+(defmethod commit :git [project & args]
```
technomancy: Can we do this in a way where each arity is spelled out instead of using `&`? I'd rather give a "wrong number of arguments" message instead of discarding extra args silently.

Other than that this looks great; thanks!

tobias:
I pushed an update that implements multi-arity for `commit` and modifies the `:none` implementation from throwing when given an argument.

technomancy:
Perfect; thanks!

Issue #2302, opened 2017-08-18
doctoc comments are showing when running "lein help faq", etc


technomancy:

technomancy:
```
~ $ lein help faq
<!-- START doctoc generated TOC please keep comment here to allow auto update -->
<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
**Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*

- [FAQ](#faq)

<!-- END doctoc generated TOC please keep comment here to allow auto update -->

# FAQ

**Q:** How do you pronounce Leiningen?  
**A:** It's LINE-ing-en. ['laɪnɪŋən]
...
```

^ This is pretty silly!

Before printing any static help text in `src/leiningen/help.clj`, we should check to see if the output includes `<!-- END doctoc generated TOC please keep comment here to allow auto update -->` and remove everything up to and including that before printing it.

irinarenteria:
@technomancy, I'd be happy to pick this up. I'll likely have questions for you or @ehashman, since I'm still quite fresh at Clojure.

technomancy:
@irinarenteria Great! Let me know if you have any questions; I'm also available in the `#leiningen` or `#clojure` channel on freenode. Thanks.

technomancy:
I'm hoping to cut a release candidate soon; if you think you could do this by the end of the week that would be great, but if not that's fine; I can fix it myself. Thanks!

ehashman:
I'm going to try to help @irinarenteria with this today!

irinarenteria:
Sorry, @technomancy, barely got around to this today (and for some reason I didn't get a notification for your comment). Will have something up by the end of this week! Spending some time with @ehashman today making sure my setup is ok.

PR #2313, opened 2017-09-16
Remove doctoc comments from static help text

irinarenteria:
Addresses issue #2302.

This is a rough first go. Any thoughts on splitting the text on a matched pattern? Since these are private functions, I haven't added any tests, but if they're required let me know. I did do a check on the `faq` file to make sure the fix works and it seems to be fine.

### DIFF
--- a/src/leiningen/help.clj
+++ b/src/leiningen/help.clj
@@ -66,9 +66,18 @@
     (some #(if (= (symbol subtask-name) (:name (meta %))) %)
           (:subtasks (meta task)))))
 
+(defn- clean-static-help
+  "Returns a string containing help content. Removes doctoc comments if they
+  are present."
+  [help-text]
+   (let [doctoc-text #"<!-- END doctoc generated TOC please keep comment here to allow auto update -->"]
+     (if (boolean (re-find doctoc-text help-text))
+       (string/triml (second (string/split help-text doctoc-text))))
+       help-text))
+
 (defn- static-help [name]
   (if-let [resource (io/resource (format "leiningen/help/%s" name))]
-    (slurp resource)))
+    (clean-static-help (slurp resource))))
 
 (declare help-for)
 
--- a/src/leiningen/help.clj
+++ b/src/leiningen/help.clj
@@ -70,10 +70,10 @@
   "Returns a string containing help content. Removes doctoc comments if they
   are present."
   [help-text]
-   (let [doctoc-text #"<!-- END doctoc generated TOC please keep comment here to allow auto update -->"]
-     (if (boolean (re-find doctoc-text help-text))
-       (string/triml (second (string/split help-text doctoc-text))))
-       help-text))
+   (let [doctoc-text "<!-- END doctoc generated TOC please keep comment here to allow auto update -->"]
+     (if (string/includes? help-text doctoc-text)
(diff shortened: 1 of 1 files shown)


src/leiningen/help.clj
```diff
@@ -66,9 +66,18 @@
     (some #(if (= (symbol subtask-name) (:name (meta %))) %)
           (:subtasks (meta task)))))
 
+(defn- clean-static-help
+  "Returns a string containing help content. Removes doctoc comments if they
+  are present."
+  [help-text]
+   (let [doctoc-text #"<!-- END doctoc generated TOC please keep comment here to allow auto update -->"]
+     (if (boolean (re-find doctoc-text help-text))
```
danielcompton: We could use `clojure.string/includes?` here possibly.

danielcompton:
This seems like a reasonable approach to take, rather than trying to fully parse the markdown. It would strip too much text if we ever moved the table of contents further down the page, but I'm not sure if that's worth worrying about.

winks:
> It would strip too much text if we ever moved the table of contents further down the page,

Now that could be easily fixed by reversing the call und using `first`  before `<!-- START doctoc...`, right?

danielcompton:
@winks, Sorry that wasn't very clear, what I'm imagining is if the doc page was changed to something like:

```md
# Tutorial

<!-- START doctoc generated TOC please keep comment here to allow auto update -->
  <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
  **Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*
...
<!-- END doctoc generated TOC please keep comment here to allow auto update -->

Leiningen is for automating Clojure projects without setting your hair on fire.
...
```

To do that, the TOC section would need to be selected and removed from the text, and then the surrounding two parts spliced together. I don't think reversing the call would be enough. I don't know if it's worth the effort and more complicated code to handle that, I'll let a core maintainer weigh in on that.

winks:
@danielcompton oops. that was exactly what I tried to say. a logically reversed call in addition ;)

technomancy:
> Since these are private functions, I haven't added any tests, but if they're required let me know. I did do a check on the faq file to make sure the fix works and it seems to be fine.

No need to add a test for this; manual testing is totally OK. This is not code that changes frequently.

This is great. The only change I would like to see is using something other than `re-find` because we're looking for an exact match rather than a regular expression pattern. `clojure.string/includes?` would be perfect.

Thanks!

irinarenteria:
@danielcompton @winks: I agree, it would be better to ensure only the exact DocToc portion is removed regardless of content structure. It might not be worth implementing at this point though, since it seems like the help content doesn’t change frequently. 

@technomancy  That makes sense. I updated that now. Funnily enough, I had to use `re-pattern` because `split` only takes a regex. Probably an indication that this is likely not the most idiomatic Clojure approach.

src/leiningen/help.clj
irinarenteria: And that eliminates the need for `boolean` too.

technomancy:
Great; thanks! Merging now.

irinarenteria:
Thank you, @technomancy!

ehashman:
:tada:

PR #2315, opened 2017-09-19
Infer default values from .git/config when :scm :git is used

### LINKED ISSUE #2274: When using `:scm {:name "git"}` pom task doesn't auto-populate the scm/tag node
From [a question in the clojars mailing list](https://groups.google.com/forum/#!topic/clojars-maintainers/GWm8tbGp6E4) I realized that the reason why some projects point to `HEAD` as the release commit in clojars (the "with this commit" link) is because they have something like the following in their project.clj: `{:scm {:name "git" :url "https://github.com..."}}`.

The `:name "git"` makes the pom task to not auto-populate the :tag and other nodes in the resulting pom, instead taking those values from the :scm map in the project.clj. But it's common for projects to only have `:name` and `:url` in that map, so `:tag` is not present in their poms, resulting in clojars pointing that link to HEAD (because that's what maven-model's `getTag` returns when it can't find an scm/tag node).

See for example [this lein-ring release](https://clojars.org/lein-ring/versions/0.12.0) where the "with this commit" link points to https://github.com/weavejester/lein-ring/commit/HEAD. Other projects where I've seen this happening are [lein-figwheel](https://clojars.org/lein-figwheel/versions/0.5.10) and [devcards](https://clojars.org/devcards/versions/0.2.3).

Even though [the sample.project.clj explains this behavior](https://github.com/technomancy/leiningen/blob/e829c12d5147ad9b5f1f50032ac1d763e20932e1/sample.project.clj#L469-L471), it seems it takes users by surprise, and also it's not clear what the implications are without a deep knowledge of multiple things.

I'm not sure what's the best way to "fix" this, but here are 2 ideas:

1. Add better documentation about the implications of having `{:scm {:name "git"}}`

All I was able to find about `:scm` in the docs (apart from the comment in sample.project.clj) is the following in DEPLOY.md (and it doesn't help much):

```
4. Conform to Sonatype OSS' requirements for uploaded artifacts' `pom.xml` files;
all you need to do is make sure the following slots are populated
properly in your `project.clj`:

  :description
  :url
  :license
  :scm
  :pom-addition
```

2. Auto-populate from git what's not defined under `:scm`

The pom task could do something like the following in [write-scm-tag](https://github.com/technomancy/leiningen/blob/e829c12d5147ad9b5f1f50032ac1d763e20932e1/src/leiningen/pom.clj#L130) when `(= "git" scm)`:

```clojure
(merge (make-git-scm ...)
       (select-keys (:scm project) ...))
```

This way projects that define any of `[:tag :url :connection :developerConnection]` would keep having those in the generated pom, those who don't will have it populated automatically, like when the `:scm` entry is not present or when it has `:name "auto"`.

Even though the change is not backwards compatible, it doesn't seem to me like it would impact negatively in many projects or any at all.

nberger:
Fixes #2274 

Implements the merge of default values inferred from `.git/config` with the values defined under the `:scm` map when scm is "git".

Refactors `write-scm-tag` a little bit so the 3 cases of scm ("auto", "git", :else) return a map that is converted with `xmlify` and `xml-tags`. I think this way is more consistent and it made it easier to implement the merge for `:scm :git`.

@technomancy: what's not implemented here (at least not yet) is the `:replace` metadata, mostly because of the following:

Is it really necessary? I understand the idea is to provide a way to use `:scm :git` while ignoring the values inferred from `.git/config`, the same way as it currently works. Let's say someone would like to not have a `:tag` node in their pom, that can be done by adding `:tag ""` in the :scm map. The same goes for the other keys (like `:connection` and `:developerConnection`). I can see that adding `:replace` metadata requires less keystrokes, but it also makes the code and doc more complex.

Also I'm thinking that the `:replace` metadata can be added later, if people ask for it. This change is not going to be 100% backwards compatible anyways.

Please let me know what you think.

### DIFF
--- a/src/leiningen/pom.clj
+++ b/src/leiningen/pom.clj
@@ -134,6 +134,8 @@
   [scm project]
   (->> (case scm
          "auto" (make-git-scm-map (resolve-git-dir project))
+         "git" (merge (make-git-scm-map (resolve-git-dir project))
+                      (make-project-scm-map project))
          ; else
          (make-project-scm-map project))
        xmlify
--- a/src/leiningen/pom.clj
+++ b/src/leiningen/pom.clj
@@ -86,19 +86,15 @@
      :dev-clone (str "ssh://git@github.com/" user "/" repo ".git")
      :browse (str "https://github.com/" user "/" repo)}))
 
-(defn- make-git-scm [git-dir]
+(defn- make-git-scm-map [git-dir]
   (try
     (let [origin (read-git-origin git-dir)
           head (read-git-head git-dir)
           urls (github-urls origin)]
-      [:scm
-       (and (:public-clone urls)
-            [:connection (str "scm:git:" (:public-clone urls))])
-       (and (:dev-clone urls)
-            [:developerConnection (str "scm:git:" (:dev-clone urls))])
-       (and head
-            [:tag head])
-       [:url (:browse urls)]])
+      (cond-> {:url (:browse urls)}
+        (:public-clone urls) (assoc :connection (str "scm:git:" (:public-clone urls)))
+        (:dev-clone urls)    (assoc :developerConnection (str "scm:git:" (:dev-clone urls)))
+        head                 (assoc :tag head)))
     (catch java.io.FileNotFoundException _)))
 
 (def disclaimer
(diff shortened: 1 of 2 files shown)


technomancy:
Sorry for the delay on this. It looks good; thanks for submitting it.

nberger:
Awesome, thanks for merging!

PR #2319, opened 2017-10-04
Replace bootclasspath with -Xverify:none

ghadishayban:
Java 9 currently fails on leiningen because Clojure 1.8.0 (latest stable) cannot reside on the BootClassPath. Rather than fixing that or waiting until Clojure 1.9.0, let's disable bytecode verification (which was the source of the startup improvement anyways.)

This allows `lein` to preserve its fast startup, without having to probe java versions.

Fixes technomancy/leiningen#2149

### DIFF
--- a/bin/lein
+++ b/bin/lein
@@ -201,8 +201,6 @@ if [ -r "$BIN_DIR/../src/leiningen/version.clj" ]; then
 else # Not running from a checkout
     add_path CLASSPATH "$LEIN_JAR"
 
-    BOOTCLASSPATH="-Xbootclasspath/a:$LEIN_JAR"
-
     if [ ! -r "$LEIN_JAR" -a "$1" != "self-install" ]; then
         self_install
     fi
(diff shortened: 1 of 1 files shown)


puredanger:
The default verify mode is -Xverify:remote, which already skips verification on classes loaded by the boot classloader. So, putting more stuff on the bootclasspath is actually *opening* security holes by putting more (unverified) bytecode on your classpath (should really do -Xverify:all to avoid that). This seems like the wrong direction to me.

ghadishayban:
closing -- issue still remains in that `lein` is about to be released broken on JDK9.

technomancy:
I don't see this as a security problem. If an attacker can affect the jar files loaded by Leiningen, bytecode verification is not going to do a lick of good; it's already a lost cause.

I'll need to run the numbers; if the same speed benefits from using the bootclasspath can be achieved this way then it's an option.

technomancy:
To be clear, this is Leiningen's own JVM which we're disabling bytecode verification for. Verification is already off for Leiningen itself and all its dependencies; this would only change verification status for plugins.

technomancy:
Based on a straw poll in IRC, most machines get plenty of benefit from `-Xverify:none` instead of the bootclasspath hack. On my personal laptop (core2duo) it takes a pretty significant performance hit; around 800ms. But this seems to be an extreme outlier, and it's possible to re-enable the bootclasspath using `LEIN_JVM_OPTS` if you know you're not going to run on Java 9, so this seems pretty reasonable while we wait for the compatibility fix from within Clojure itself.

Issue #2321, opened 2017-10-18
Leiningen 2.8.0 requires Java 1.8.0u131 by default


brixtonasias:
When trying to run a build with a Java version lower than `1.8.0u131` builds fail with the message: `Unrecognized VM option 'UseCGroupMemoryLimitForHeap'`

The website tells users that: 

> Make sure you have a Java JDK version 6 or later.

The fix in the FAQ to use `:jvm-opts ^:replace []` in `project.clj` results in:

```
Exception in thread "main" java.lang.RuntimeException: No reader function for tag object, compiling:(/private/var/folders/hl/wc_8mc8950dbbp5pdnz56br00000gn/T/form-init1787191864903406657.clj:1:8328)
```

What is the actual solution to this when one can not / does not want to upgrade the Java version?

acron0:
```clojure
(def ^:private cgroups-jvm-opts
  ;; this assumes the JVM version Leiningen is run under matches the project
  (let [[v u] (re-seq #"[^-_]+" (System/getProperty "java.runtime.version"))]
    (if (or (and (= v "1.8.0") (>= (Integer. u) 131))
            (not= v "1.7.0")
            (not= v "1.6.0"))
      ["-XX:+UnlockExperimentalVMOptions" "-XX:+UseCGroupMemoryLimitForHeap"])))
```
This code is wrong. When `v=1.8.0` and `u<131` the `or` will return `true` due to the `not=` clauses.

```clojure
user> (let [[v u] ["1.8.0" "111"]]
        (or (and (= v "1.8.0") (>= (Integer. "111") 131))
            (not= v "1.7.0")
            (not= v "1.6.0")))
true
```
I'm curious as to the `not=` clauses and why they were included.

technomancy:
It looks like this is really two bugs; one detecting the JVM version and one preventing the manual override from working. 

Can you gist the contents of `/private/var/folders/hl/wc_8mc8950dbbp5pdnz56br00000gn/T/form-init1787191864903406657.clj` so we can address the latter?

Also a great reminder to update the web site to recommend a version of the JDK that still gets security updates.

scstarkey:
I have it even worse. I'm using the latest Java 8. Here's my environment (with no changes to `project.clj`):

`$ lein version`
`Leiningen 2.8.0 on Java 1.8.0_152 Java HotSpot(TM) 64-Bit Server VM`

[...]

PR #2323, opened 2017-10-18
Fix boolean logic in Java version check for cgroups memory limit JVM option.

acron0:
This was written about in https://github.com/technomancy/leiningen/issues/2321

There's an issue in the code which means Java versions that don't support `UseCGroupMemoryLimitForHeap` are still having it applied.


acron0:
Fixing whitespace commits

PR #2324, opened 2017-10-18
Fix boolean logic in Java version check for cgroups memory limit JVM option

acron0:
This was written about in https://github.com/technomancy/leiningen/issues/2321

There's an issue in the code which means Java versions that don't support `UseCGroupMemoryLimitForHeap` are still having it applied.

### DIFF
--- a/leiningen-core/src/leiningen/core/project.clj
+++ b/leiningen-core/src/leiningen/core/project.clj
@@ -523,13 +523,16 @@
 
 ;; give reasonable -Xmx defaults when containerized, if JVM is new enough
 ;; https://blogs.oracle.com/java-platform-group/java-se-support-for-docker-cpu-and-memory-limits
+(defn use-cgroups-memory-limit-for-heap?  [version]
+  (let [[v u] (re-seq #"[^-_]+" version)
+        v' (Integer. (apply str (re-seq #"\d" v)))
+        u' (Integer. u)]
+    (and (>= v' 180) (>= u' 131))))
+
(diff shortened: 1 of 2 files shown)


acron0:
Should I be worried about this? 
```
ERROR in (test-b) (test_b.clj:6)
Uncaught exception in test fixture
expected: nil
  actual: java.lang.Exception: Don't panic. This is an expected exception.
```

technomancy:
Thanks. Unfortunately I don't think making these integers is the right choice, since I read that Java 9 will start returning "9" as its version instead of "1.9.0". Also, it's using `when` for the return value instead of side-effects.

acron0:
Thanks for the feedback. Are you happy with a straight string comparison then, like in the original? Also I don't quite understand your closing remark. It's possible this is an idiom I am unaware of. The original code used a single-branch `if` which I almost automatically always replace with a `when` - but are the semantics of `when` that it's for side-effects?

technomancy:
Yes, I think the string comparison is the way to go.

> are the semantics of when that it's for side-effects?

Right; this is why `when` has an implicit `do` in the macroexpansion. Replacing single-branched `if` with `when` only makes sense in cases where it's done for side-effects.

technomancy:
Perfect; thanks!

Meidan:
is there a built version with this patch?

Meidan:
I couldn't find any, so I suggest reverting the stable branch to 2.7.1 until a new release is available

PR #2330, opened 2017-10-19
Don't use UseCGroupMemoryLimitForHeap on OS X

### LINKED ISSUE #2329: 2.8.0 on macos - error related to +UseCGroupMemoryLimitForHeap
On macos High Sierra, Java 1.8.0_151-b12, Leiningen 2.8.0 I see warnings:
```
Java HotSpot(TM) 64-Bit Server VM warning: Unable to open cgroup memory 
limit file /sys/fs/cgroup/memory/memory.limit_in_bytes (No such file or directory)
```
I know about Issue #2321 that has been fixed but this is not quite the same; my JRE does have the `+UseCGroupMemoryLimitForHeap` option but macos doesn't have cgroups so the warning is printed. It is only a warning but still would be nice if it were fixed. I got my leiningen from Homebrew.

Steps to reproduce:
```
lein new app xyz
cd xyz
lein run
```

winks:
Trying to fix #2329 - untested for lack of access to a Mac right now.

### DIFF
--- a/leiningen-core/src/leiningen/core/project.clj
+++ b/leiningen-core/src/leiningen/core/project.clj
@@ -523,13 +523,13 @@
 
 ;; give reasonable -Xmx defaults when containerized, if JVM is new enough
 ;; https://blogs.oracle.com/java-platform-group/java-se-support-for-docker-cpu-and-memory-limits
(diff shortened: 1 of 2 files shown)


technomancy:
Thanks; looks good once the tests are fixed.

PR #2343, opened 2017-10-26
Fix the cgroups fix, again

winks:
Proposition a) Try to fix the java version detection for cgroups flags

xref #2344 - choose one


winks:
Closing this, as it was an alternative scenario for #2344 which was merged.

Issue #2342, opened 2017-10-26
lein-pkg obscures exit code unless trampolining


bhb:
This behavior changed from 2.7.1 to 2.8.0

Repro:

1. Add the following function in any namespace - in this example, it'll be `foo.core`

```clojure
(defn -main []
  (System/exit 1))
```
2. `lein run -m foo.core/-main; echo $?`

Expected: The exit status of lein should be 1
Actual: The exit status of lein is 0

If I run `lein trampoline run -m foobar.core/-main; echo $?`, lein will exit 1.

Why would someone want this behavior? In our case, we run `cljsbuild` with a special warning handler that exits 1 if there are warnings. We run this check in CI. With lein 2.7.1, a warning would fail the compile and therefore fail the build. In lein 2.8.0, the compile step passes.

technomancy:
This doesn't happen when I try it; can you provide a repro case?

bhb:
Hm, sorry about that. I thought this simplified repro (simpler than the clsjbuild repro) worked, but now I can't reproduce either. I'll try to find another repro involving a subprocess.

bhb:
This is very weird - now I can't even repro the original issue. I have a lot of the original bug in our CI output, and I could repro locally yesterday on 2.8.0, but no more.

I'm closing this for now, but I'll reopen if I can find a consistent repro. Sorry for the noise.

bhb:
Ah, I discovered the source of my confusion - this only reproduces if I install `lein` with homebrew. In the example below, `lein` is installed with homebrew, but `lein2` is installed using the lein script (from [these instructions](https://leiningen.org/))

```
Distillers-Mac-3:foo distiller$ which lein
/usr/local/bin/lein
Distillers-Mac-3:foo distiller$ lein -v
Leiningen 2.8.0 on Java 1.8.0_131 Java HotSpot(TM) 64-Bit Server VM
Distillers-Mac-3:foo distiller$ lein run -m foo.core/-main; echo $?
Java HotSpot(TM) 64-Bit Server VM warning: Unable to open cgroup memory limit file /sys/fs/cgroup/memory/memory.limit_in_bytes (No such file or directory)
0
Distillers-Mac-3:foo distiller$ ./lein2 -v
Leiningen 2.8.0 on Java 1.8.0_131 Java HotSpot(TM) 64-Bit Server VM
Distillers-Mac-3:foo distiller$ ./lein2 run -m foo.core/-main; echo $?
[...]

PR #2344, opened 2017-10-26
Get rid of cgroups special handling in core

winks:
Proposition b) Remove the java version detection for cgroups flags

xref #2343 - choose one


technomancy:
Thanks. Seeing #2342 makes me want to go with the simpler solution here and just get 2.8.1 out asap.

PR #2349, opened 2017-10-31
Add `init-ns` to default template, fixes #2350

### LINKED ISSUE #2350: It would be nice if the default template had a value for `:repl-options {:init-ns }`
I generally do `lein new foo` whenever I start a new project. A subsequent `lein repl` puts me in 
the default ns instead of `foo.core`

slipset:


technomancy:
Nice; thanks.

PR #2354, opened 2017-11-06
Add description for changing maven read timeout

sashton:
This is a documentation change to describe how to increase the read timeout for the Maven Wagon transport.


technomancy:
Nice; thanks.

PR #2362, opened 2017-11-27
Avoid "You have $CLASSPATH set" errors on upgrade

### LINKED ISSUE #2361: `lein upgrade` with bash script throws accidental warning about $CLASSPATH
@cswl reports via IRC that when a user runs `lein upgrade`, they encounter a warning:

```
WARNING: You have $CLASSPATH set, probably by accident.
It is strongly recommended to unset this before proceeding.
```

This appears to be because we `exec` the new script once downloaded:

https://github.com/technomancy/leiningen/blob/4169f26ca83e0ed70c70a415e45d673acade3c4d/bin/lein#L281-L288

Since we have already exported the classpath within the currently running lein script, we still use the same environment with classpath to `exec` the new script, hence the warning. Perhaps we could unset the CLASSPATH variable before `exec`ing the newly downloaded script to avoid this.

ehashman:
Fixes #2361.

When we exec the newly downloaded leiningen script, it will replace our
current shell process and recalculate the classpath for us, so we can
safely unset this here. This avoids the new script throwing warnings
about the CLASSPATH variable being set (from the previous run) on
upgrades.

### DIFF
--- a/bin/lein
+++ b/bin/lein
@@ -285,6 +285,7 @@ elif [ "$1" = "upgrade" ] || [ "$1" = "downgrade" ]; then
                         echo "Leiningen is already up-to-date."
                     fi
                     mv "$TARGET" "$SCRIPT" && chmod +x "$SCRIPT"
+                    unset CLASSPATH
                     exec "$SCRIPT" version
                 else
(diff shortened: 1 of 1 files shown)


technomancy:
Thanks for the fix!

PR #2363, opened 2017-11-28
Port CLASSPATH check to issue warning to lein.bat

winks:
Related to #2361

### DIFF
--- a/bin/lein.bat
+++ b/bin/lein.bat
@@ -10,6 +10,11 @@ if "%LEIN_VERSION:~-9%" == "-SNAPSHOT" (
     set SNAPSHOT=NO
 )
 
+if not "x%CLASSPATH%" == "x" (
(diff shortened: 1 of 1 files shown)


technomancy:
Oh, I just took a look at `bin/lein.bat` and noticed that it appears to overwrite the value of `CLASSPATH` rather than appending to it. In that case the bad behavior we are trying to warn against will never happen, so... I think the warning isn't needed?

winks:
Right, I missed that. So this can probable be discarded.

Bit unhappy of the overall state of everything being different. Something that switching to Powershell won't fix either. Ah, if only we could write `lein/lein.bat` in Clojure and transpile to shell/batch :P

technomancy:
I don't see this as a bad thing that they are different; overall the fact that the batch file isn't saddled with the same historical baggage as the bash script is probably a plus.

PR #2366, opened 2017-12-08
[docs] Update TEMPLATES.md to refer to resources dir rather than src dir

venantius:
The current documentation claims template files live in `src`, rather than
`resources` - a fact trivially disprovable by either (a) referring to
Leiningen's default templates in `resources/leiningen/new` or (b) actually
trying to make a new template yourself.

This commit updates the documentation to refer to the correct directory.


technomancy:
Good catch; thanks.

PR #2379, opened 2017-12-21
Add project coordinate data to JAR metadata

conormcd:
This causes entries like:

```
Leiningen-Project-ArtifactId: bar
Leiningen-Project-GroupId: foo
Leiningen-Project-Version: 1.2.3-SNAPSHOT
```

to be added to `META-INF/MANIFEST.MF` in JARS for projects with `project.clj` files like:

```clojure
(defproject foo/bar "1.2.3-SNAPSHOT"
  ...
  )
```

### DIFF
--- a/src/leiningen/jar.clj
+++ b/src/leiningen/jar.clj
@@ -22,10 +22,13 @@
 (defn- unix-path [path]
   (.replace path "\\" "/"))
 
-(def ^:private default-manifest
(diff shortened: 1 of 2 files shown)


technomancy:
Following up with a conversation; this seems redundant with `pom.properties` but this one is useful when you want to know the name of the current application when running in an uberjar; you can't get that info with the existing metadata we emit.

PR #2386, opened 2018-01-11
Fix Gitlab mirror

tirkarthi:


technomancy:
Thanks, but this is fixed on the master branch. We don't push to the stable branch unless it's for a release.

Issue #2380, opened 2017-12-22
Tab completion with Leiningen REPL after typing a period in a namespace triggers an NPE


venantius:
Bit of an oddly specific bug, I know.

TL;DR - if I type `clojure.co` and then try to tab-complete, an NPE is triggered. Tab-completion works fine if I hit tab at `cloj` or at `clojure.core/`, but after that first period and before the slash I'll get NPE'd. I ran into this on a project and then was able to reproduce outside of a project context. 

Now the *expected* behavior would be for tab completion to work the way it did when I hit tab at `cloj`, i.e. to show me the list of namespaces that are potential results for what I've typed so far.

Full command history and stacktrace:

```ursa@ziz:~
▸ lein repl
nREPL server started on port 55883 on host 127.0.0.1 - nrepl://127.0.0.1:55883
REPL-y 0.3.7, nREPL 0.2.12
Clojure 1.8.0
Java HotSpot(TM) 64-Bit Server VM 9.0.1+11
    Docs: (doc function-name-here)
          (find-doc "part-of-name-here")
  Source: (source function-name-here)
 Javadoc: (javadoc java-object-or-class-here)
    Exit: Control+D or (exit) or (quit)
 Results: Stored in vars *1, *2, *3, an exception in *e

user=> clojure.corNullPointerException   complete.core/fn--632/iter--624--633/fn--634 (core.clj:66)


CompilerException java.lang.ClassNotFoundException: clojure.cor, compiling:(null:0:0)
user=> (pst)
clojure.lang.Compiler$CompilerException: java.lang.ClassNotFoundException: clojure.cor, compiling:(null:0:0)
                                Compiler.java:6688 clojure.lang.Compiler.analyze
                                Compiler.java:6625 clojure.lang.Compiler.analyze
                                Compiler.java:6931 clojure.lang.Compiler.eval
                                Compiler.java:6890 clojure.lang.Compiler.eval
                                     core.clj:3105 clojure.core/eval
                                     core.clj:3101 clojure.core/eval
                                      main.clj:240 clojure.main/repl[fn]
                                      main.clj:240 clojure.main/repl[fn]
                                      main.clj:258 clojure.main/repl[fn]
                                      main.clj:258 clojure.main/repl
                                      main.clj:174 clojure.main/repl
                                  RestFn.java:1523 clojure.lang.RestFn.invoke
                         interruptible_eval.clj:87 clojure.tools.nrepl.middleware.interruptible-eval/evaluate[fn]
[...]

PR #2397, opened 2018-02-05
Bump reply to 0.3.8

WillChilds-Klein:
closes #1390

PR #2398, opened 2018-02-05
Bump version of reply and clojure-complete

tirkarthi:
This PR closes https://github.com/technomancy/leiningen/issues/2380 . clojure-complete "0.2.5" is released with the fix and reply is also released with the updated clojure-complete library as "0.3.8". This PR bumps the version of clojure-complete and reply. Tested it locally and it's working fine now.

Just saw a PR (https://github.com/technomancy/leiningen/pull/2397) that only upgrades reply version. I tested that PR and it still causes the NPE. Though clojure-complete is not used directly I don't know why an older version causes error with reply upgraded. have upgraded both and let me know if there is any place I have missed.

cc @venantius


venantius:
Nice turnaround! :) 👍

technomancy:
Looks good; thanks.

Issue #2358, opened 2017-11-21
EPL-2.0 and Leiningen


danielcompton:
Recently, the Eclipse Foundation have released the [EPL-2.0](https://www.eclipse.org/legal/epl-2.0/). There are a few improvements from the EPL-1.0 that Clojure and most Clojure projects use:

> 1.4. What are the major changes between EPL-1.0 and EPL-2.0?
> * The term of art is now referred to as "module" rather than "file";
> * The choice of law provisions has been removed;
> * The license is now suitable for scripting languages such as JavaScript; and
> * The license now includes an option to add a secondary license for GPL-2.0+ compatibility.

Potentially the most interesting to many people is the ability to add a secondary license to allow GPL-2.0+ compatibility.

I've had some discussions on the EPL-discuss mailing list: http://dev.eclipse.org/mhonarc/lists//epl-discuss/msg00158.html, and there is much more detail in the [EPL-2.0 FAQ](https://www.eclipse.org/org/documents/epl-2.0/faq.php#h.hz76esowpykz). I would encourage anyone interested in this topic to read the FAQ.

I've put a few abridged questions and answers that I think will answer people's most common questions. You should read the FAQ for the full context on this though.

Q: [What is the impact of adding a secondary GPL license?](https://www.eclipse.org/org/documents/epl-2.0/faq.php#h.lgjcpvoq08a9)
A: It is almost the same as dual licensing EPL/GPL

Q: [How can I upgrade a project from EPL-1.0 to EPL-2.0?](https://www.eclipse.org/org/documents/epl-2.0/faq.php#h.tci84nlsqpgw)
A: The EPL allows new versions of the license to be adopted by projects with little work. A project can use the new version by simply updating the file headers and notices. It would be good practice to discuss this before relicensing though.

Q: [How can I upgrade a project to EPL-2.0 with GPL secondary license?](https://www.eclipse.org/org/documents/epl-2.0/faq.php#h.cktcxkzh8ks4)
A: You must gain permission from all copyright holders to re-license the content (Ed: Projects with a CLA would allow you to do this, as you have already gotten approval from contributors.)

Q: [Some organizations do not use GPL-licensed content. What are the impacts of licensing a project with EPL-2.0 + GPL secondary license?](https://www.eclipse.org/org/documents/epl-2.0/faq.php#h.ofnb7ks8rr0y)
[...]

PR #2403, opened 2018-02-19
Switch to EPL-2.0 in `lein new` templates

yegortimoshenko:
GitHub now properly identifies the license: https://github.com/yegortimoshenko/leiningen-epl-2.0-sandbox

See #2358. Some notes:

* this doesn't change license on Leiningen or any of its components;

* this does not include GPLv3 dual licensing (I believe that would constitute a breaking change);

* license file matches https://www.eclipse.org/org/documents/epl-2.0/EPL-2.0.txt bit-by-bit;

* license name now uses the SPDX identifier, which is seemingly less vague than previous "Eclipse Public License".

### DIFF
--- a/resources/leiningen/new/app/LICENSE
+++ b/resources/leiningen/new/app/LICENSE
@@ -1,214 +1,277 @@
-THE ACCOMPANYING PROGRAM IS PROVIDED UNDER THE TERMS OF THIS ECLIPSE PUBLIC
-LICENSE ("AGREEMENT"). ANY USE, REPRODUCTION OR DISTRIBUTION OF THE PROGRAM
-CONSTITUTES RECIPIENT'S ACCEPTANCE OF THIS AGREEMENT.
+Eclipse Public License - v 2.0
+
+    THE ACCOMPANYING PROGRAM IS PROVIDED UNDER THE TERMS OF THIS ECLIPSE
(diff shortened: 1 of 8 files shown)


technomancy:
OK, I merged this and also fixed the template readmes and included the secondary license in 694691f.

yegortimoshenko:
OK!

Also, while I think GPLv2 is closer to the spirit of EPL (due to lack of tivoization clause), it's also problematic and obsolete (mostly what bothers me is that there is no grace period on violations, and torrenting executables is technically prohibited).

Perhaps GPLv3 with classpath exception, or even LGPLv3 would be a better idea? I certainly think (L)GPLv3 is the better license, and I think you would agree.

technomancy:
I agree that GPLv3 is better, but the provision to allow 2.0 or later is embedded inside the EPL itself, not just in the SPL license identifier. So I don't think we can change that without forking the EPL.

yegortimoshenko:
Definition from EPL-2.0:

> “Secondary License” means either the GNU General Public License, Version 2.0, or any later versions of that license, including any exceptions or additional permissions as identified by the initial Contributor. 

It even seems that GPLv2-only is not a valid secondary license, given this definition and a related FAQ entry: https://www.eclipse.org/legal/epl-2.0/faq.php#h.lza2unrion3b

However, LGPL is also an invalid secondary license.

PR #2404, opened 2018-02-20
Use GPL-2.0-or-later as secondary license

yegortimoshenko:
See: https://github.com/technomancy/leiningen/pull/2403#issuecomment-367117859

Other minor changes:

* Use FSF-suggested wording in license description (added "as
  published by the Free Software Foundation")
* Update README.md and project.clj in resources/leiningen/template

### DIFF
--- a/resources/leiningen/new/app/README.md
+++ b/resources/leiningen/new/app/README.md
@@ -38,6 +38,7 @@ http://www.eclipse.org/legal/epl-2.0.
 
 This Source Code may also be made available under the following Secondary
 Licenses when the conditions for such availability set forth in the Eclipse
-Public License, v. 2.0 are satisfied: GNU General Public License, version 2
-with the GNU Classpath Exception which is
-available at https://www.gnu.org/software/classpath/license.html.
+Public License, v. 2.0 are satisfied: GNU General Public License as published by
+the Free Software Foundation, either version 2 of the License, or (at your
+option) any later version, with the GNU Classpath Exception which is available
+at https://www.gnu.org/software/classpath/license.html.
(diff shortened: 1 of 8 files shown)


technomancy:
This looks good, but I'm not 100% sure it's a good idea to replace the license identifier in `project.clj` as the current one is the one recommended by the Eclipse foundation. Their recommendation is to use the "or later version" clause, but their recommendation for what to put in the pom does not reflect this.

@pmonks You are pretty knowledgeable about SPDX license identifiers; could you chime in on this?

pmonks:
@technomancy you flatter me.  I like it.  ;-)

But FWIW the SPDX committee deprecated [`GPL-2.0+` syntax](https://spdx.org/licenses/GPL-2.0+.html) in favour of [`GPL-2.0-or-later` syntax](https://spdx.org/licenses/GPL-2.0-or-later.html) a year or two ago, so the SPDX expression in the PR looks good to me (assuming you're ok with the semantic change from "exactly GPL v2.0 only" to "GPL v2.0, or any later version").

What I'm not entirely sure of (but I'm sure you are!) is whether `:license :name` is supposed to contain an SPDX expression, and whether that might have adverse effects in downstream non-Leiningen consumers of a generated `pom.xml` (i.e. Maven projects).  I can't imagine how it would matter, but I'm by no means a Maven expert so I don't know if/how this property might be consumed by Maven and its menagerie of plugins.

technomancy:
@pmonks The relevant document in this case is Maven's pom reference, since this element is emitted straight to the pom files which are deployed to the repositories and not anywhere else: https://maven.apache.org/pom.html#Licenses

I read it as saying that it's intended as a human-readable name of the license, but IMO there's no reason not to use an spdx license identifier here, since those are human-readable. It's when you make assumptions about what others put in it that you run into trouble. =)

PR #2405, opened 2018-02-25
Enforce TLS 1.2 for PowerShell downloader

D-side:
GitHub has stopped supporting TLS < 1.2 just a few days ago:
https://github.com/blog/2507-weak-cryptographic-standards-removed
...as I found out yesterday attempting to install Leiningen 😐

On up-to-date Windows 10 PowerShell is configured to only use TLS 1.0 and SSL 3 (`Tls, Ssl3` from [here](https://msdn.microsoft.com/ru-ru/library/system.net.securityprotocoltype(v=vs.110).aspx)).

With this change, installation succeeds.

From what I understand, tests to not cover installation scripts and thus don't need any updates.

### DIFF
--- a/bin/lein.bat
+++ b/bin/lein.bat
@@ -154,7 +154,7 @@ if "x%HTTP_CLIENT%" == "x" goto TRY_POWERSHELL
 call powershell -? >nul 2>&1
 if NOT ERRORLEVEL 0 goto TRY_WGET
     set LAST_HTTP_CLIENT=powershell
-    powershell -Command "& {param($a,$f) $client = New-Object System.Net.WebClient;  $client.Proxy.Credentials =[System.Net.CredentialCache]::DefaultNetworkCredentials; $client.DownloadFile($a, $f)}" ""%2"" ""%1""
+    powershell -Command "& {param($a,$f) $client = New-Object System.Net.WebClient; [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12; $client.Proxy.Credentials =[System.Net.CredentialCache]::DefaultNetworkCredentials; $client.DownloadFile($a, $f)}" ""%2"" ""%1""
     SET RC=%ERRORLEVEL%
     goto EXITRC
 


joefromct:
FYI, i have a new windows 10 build and tried the work around in the stack overflow posted above; I still get this result:

```` bash
PS C:\Users\me> [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
PS C:\Users\me> lein self-install
Downloading Leiningen now...
Exception calling "DownloadFile" with "2" argument(s): "The request was aborted: Could not create SSL/TLS secure
channel."
At line:1 char:145
+ ... che]::DefaultNetworkCredentials; $client.DownloadFile($a, $f)} "https ...
+                                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], MethodInvocationException
    + FullyQualifiedErrorId : WebException


Failed to download https://github.com/technomancy/leiningen/releases/download/2.8.1/leiningen-2.8.1-standalone.zi

It is possible that the download failed due to "powershell",
"curl" or "wget"'s inability to retrieve GitHub's security certificate.
The suggestions below do not check certificates, so use this only if
you understand the security implications of not doing so.

The PowerShell failed to download the latest Leiningen version.
Try to use "curl" or "wget" to download Leiningen by setting up
the HTTP_CLIENT environment variable with one of the following
values:

  a) set HTTP_CLIENT=wget --no-check-certificate -O
  b) set HTTP_CLIENT=curl -f -L -k -o

NOTE: Make sure to *not* add double quotes when setting the value
      of HTTP_CLIENT

PS C:\Users\me>
````

D-side:
@joefromct it has no effect because `lein.bat`launches PowerShell in a subprocess, so the PS session that actually performs the download is not affected by your first line.

For the workaround to work, you'll either have to make this a default for all newly opened PowerShells (no idea how or whether it's even possible), or apply the same change to your `lein.bat` as I did here.

joefromct:
ah, ok well that makes sense. 

I actually installed lein with git bash ... and that dropped the standalone.jar into the right place and i could get to a repl.

I'm sure i'll have future problems however that might require that sort of thing.  I'm still not exactly comfy on windows.

D-side:
Can't say the workaround via modifying `lein.bat` as in this PR is harder than enforcing TLS 1.2 PowerShell-wide.

---

### Make PowerShell use TLS 1.2 by default everywhere

This is, I think, the least invasive way: changes do not require admin privileges, only apply to the _current user_ (not system-wide) and have a minimal impact on security.

First, you need to [set execution policy](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.security/set-executionpolicy) _for you_ (not system-wide) to something that would allow execution of "profile" scripts, like so:

    Set-ExecutionPolicy remoteSigned -Scope CurrentUser # and answer Y in the prompt

This allows unrestricted execution of **local** PS scripts (including user-created PS profiles, more on that later) and requires scripts downloaded from the internet to be signed by a trusted publishers; this is the least permissive option after the default `restricted`. This change is persisted in Windows registry.

Then create a [_PS profile file_](https://technet.microsoft.com/en-us/library/bb613488(v=vs.85).aspx) in one of the user-scoped locations (starting with `%UserProfile%`; e. g. `%UserProfile%\My Documents\WindowsPowerShell\profile.ps1`, typically `C:\Users\<username>\My Documents\WindowsPowerShell\profile.ps1`, you may have to create folders) and edit it to contain what you want to execute on startup of every PS session. For this issue, this is how the profile should look like:

    [System.Net.ServicePointManager]::SecurityProtocol = [System.Net.SecurityProtocolType]::Tls12

---

### To roll back

Remove the created _PS profile file_, then execute this command:

    Set-ExecutionPolicy restricted -Scope CurrentUser # and answer Y in the prompt

technomancy:
Sorry, just to confirm, your recent comments indicate that there's a workaround folks can use without this patch, but applying this patch would make it unnecessary; is that correct?

D-side:
@technomancy yes, exactly.

winks:
This breaks Windows 7, see discussion in #2412

PR #2411, opened 2018-03-23
Use `clojars` repo consistently in document

rymndhng:
Ensures that the two sections of the document (configuring GPG and setting up Clojars as a default deploy repository) work together.

### DIFF
--- a/doc/DEPLOY.md
+++ b/doc/DEPLOY.md
@@ -154,7 +154,7 @@ First write your credentials map to `~/.lein/credentials.clj` like so:
 
 ```clj
 {#"blueant" {:password "locative1"}
- #"https://clojars.org/repo"
+ #"https://repo.clojars.org"
  {:username "milgrim" :password "locative1"}
  "s3p://s3-repo-bucket/releases"
  {:username "AKIAIN..." :passphrase "1TChrGK4s..."}}


doc/DEPLOY.md
```diff
@@ -154,7 +154,7 @@ First write your credentials map to `~/.lein/credentials.clj` like so:
 
 ```clj
 {#"blueant" {:password "locative1"}
- #"https://clojars.org/repo"
+ #"https://repo.clojars.org"
```
rymndhng: When this regex does not match the section below on `:deploy-repositories`, `lein release` does not pick up the credentials from this file.

danielcompton:
@tobias deploys are fine to go through `repo.clojars.org` right? Couldn't remember.

tobias:
Yes, the CDN at `repo.clojars.org` proxies to `clojars.org/repo` when it gets a `POST` or `PUT`, so it's fine to use the CDN for deploys.

rymndhng:
From both your POVs: @tobias @danielcompton, what would be the canonical URL to use? 

For a bit more context, if someone followed the docs as is and ended this config:

``` clojure
;; ~/.lein/credentials.clj.gpg
{#"https://repo.clojars.org"  {:username "milgrim" :password "locative1"}}

;; project.clj
:deploy-repositories {"releases" {:url "https://repo.clojars.org" :creds :gpg}}
```

The deploy repository does not match the credentials URL regex, and thus they'll always be prompted to enter username/password when running `lein deploy`. An experienced `lein` user would know what's going on and fix it, however a beginner may not make the link between the two.

danielcompton:
I _think_ that any new users should use `https://repo.clojars.org`. 

We keep `clojars.org/repo` around for backwards compatibility with old clients that haven't updated, and for people using old versions of Java that don't support the TLS versions that Fastly provides (and maybe also SNI, can't remember off-hand).

As a rule of thumb it's probably safe to say that `clojars.org/repo` should only be used if you have a specific reason that you can't use `repo.clojars.org`.

technomancy:
Makes sense; thanks.

Issue #2391, opened 2018-01-18
Can't run an alias with a duplicit definition


katox:
If you define an alias more than once an error occurs on invocation. 

This is a mainly problem if you add aliases in different profiles because you need to be sure that all aliases are globally unique.

A minimal testcase to reproduce this this issue is here: https://github.com/katox/lein-dup-alias-testcase

technomancy:
My bad; closed this by accident.

PR #2413, opened 2018-04-11
Ensure aliases are replaced when merging profiles

aroemers:
The simple workaround (and maybe the advised way to go?) for issue #2391 is by adding `^:replace` metadata onto the alias value inside a profile:

```clj
(defproject
  ...
  :aliases {"foo" ["bar"]}
  :profiles {:alice {:aliases {"foo" ^:replace ["baz"]}}}
  ...)
```

Yet, I cannot think of a situation where one wants the standard (meta-)merging of aliases. Users may expect above replace behaviour by default. This PR is a possible realisation of such default behaviour.

Feedback is welcome, as I do not know the entire leiningen codebase by head.

P.S. The tests are failing, but they do not seem related to this PR. It cannot find some dev-resources, such as `leiningen-core/dev-resources/p1.clj`...

### DIFF
--- a/leiningen-core/src/leiningen/core/project.clj
+++ b/leiningen-core/src/leiningen/core/project.clj
@@ -124,6 +124,15 @@
     (with-meta obj m)
     obj))
 
+(defn- vary-meta*
+  "Returns an object of the same type and value as obj, with
+  (apply f (meta obj) args) as its metadata, if the object can hold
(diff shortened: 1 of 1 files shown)


technomancy:
Sorry for the delay on this; it looks great. Thanks for submitting it.

PR #2415, opened 2018-04-13
Protect the '.' in the regex

yogsototh:
This is a minor issue, but it looks a bit safer to me.

Of course I tried it.


winks:
Thanks!

PR #2417, opened 2018-04-14
Bump commons-io version

ehashman:
Upstreaming this patch I wrote to ensure we can build against the latest `commons-io`: https://salsa.debian.org/clojure-team/leiningen-clojure/blob/master/debian/patches/0004-Bump-commons-io-version.patch


technomancy:
Makes sense, thanks.

Issue #2416, opened 2018-04-14
Installing Leiningen


Danimal66:
Hello, I am a programmer wanting to learn Clojure. I am using Windows 10 and am not quite sure how to properly install Leiningen. I saved the .bat file and added the folder it's stored in to my path in environment variables but when I go to command prompt to try and do a lein self-install it tells me that it doesn't recognize the command "lein". I was wondering if you could provide me with more assistance. I have tried to look up videos or articles explaining what to do, but I can't figure out what I am doing incorrectly. Thank you for your time!

Danimal66:
Thank you for the quick response! When I ran the .bat file and restarted it got rid of the error where it doesn't recognize "lein" but it now it pops up a long list of errors that "is not recognized as an internal or external command, operable program or batch file." I think I am at the part of the installation where it tells me to make it executable. But I'm confused by the wording. Am I supposed to make the .bat file into a .exe now?

winks:
No, that's only for the `lein` shell script under Linux/UN*X.
If you are using `cmd` then just running `lein` should work.

If you are using `PowerShell` and are in the same dir, then `.\lein.bat` or `.\lein` should work.
If you are using `PowerShell` and are not in the same dir, then just `lein` should suffice.

stevechae:
Also note that Windows install (via `lein.bat`) seems to be broken at the moment. The fix is outlined here: https://github.com/technomancy/leiningen/issues/2412

Danimal66:
I get a different error than they're receiving in the fix you provided, I am probably doing something wrong, but this is the error message I get currently:

C:\Users\danam>lein

C:\Users\danam>#!/usr/bin/env bash
'#!' is not recognized as an internal or external command,
operable program or batch file.

C:\Users\danam># Ensure this file is executable via `chmod a+x lein`, then place it
'#' is not recognized as an internal or external command,
operable program or batch file.

C:\Users\danam># somewhere on your $PATH, like ~/bin. The rest of Leiningen will be
'#' is not recognized as an internal or external command,
operable program or batch file.

C:\Users\danam># installed upon first run into the ~/.lein/self-installs directory.
'#' is not recognized as an internal or external command,
operable program or batch file.

C:\Users\danam>export LEIN_VERSION="2.8.1"
'export' is not recognized as an internal or external command,
operable program or batch file.

C:\Users\danam>case $LEIN_VERSION in
'case' is not recognized as an internal or external command,
operable program or batch file.

[...]

Issue #2412, opened 2018-04-11
lein self-install error on Windows


pedro-w:
I tried to install leiningen 2.8.1 on Windows 10 64-bit. I don't have wget or curl installed so the batch file used PowerShell to fetch the jars. I got the following error - of which I think "Could not create SSL/TLS secure channel" is the significant part.

```
>lein self-install
Downloading Leiningen now...
Exception calling "DownloadFile" with "2" argument(s): "The request was
aborted: Could not create SSL/TLS secure channel."
At line:1 char:145
+ ... che]::DefaultNetworkCredentials; $client.DownloadFile($a, $f)} "https ...
+                                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], MethodInvocationException
    + FullyQualifiedErrorId : WebException


Failed to download https://github.com/technomancy/leiningen/releases/download/2.8.1/leiningen-2.8.1-standalone.zip

It is possible that the download failed due to "powershell",
"curl" or "wget"'s inability to retrieve GitHub's security certificate.
The suggestions below do not check certificates, so use this only if
you understand the security implications of not doing so.

The PowerShell failed to download the latest Leiningen version.
Try to use "curl" or "wget" to download Leiningen by setting up
the HTTP_CLIENT environment variable with one of the following
values:

  a) set HTTP_CLIENT=wget --no-check-certificate -O
  b) set HTTP_CLIENT=curl -f -L -k -o

NOTE: Make sure to *not* add double quotes when setting the value
      of HTTP_CLIENT
```

pedro-w:
I was able to fix it by specifying the security protocol, see below. Please let me know if this is useful and I will submit it as a github PR.

```patch
--- lein.bat.original   2018-04-09 11:27:07.549354500 +0100
+++ lein.bat    2018-04-09 11:28:43.442263400 +0100
@@ -154,7 +154,7 @@
 call powershell -? >nul 2>&1
 if NOT ERRORLEVEL 0 goto TRY_WGET
     set LAST_HTTP_CLIENT=powershell
-    powershell -Command "& {param($a,$f) $client = New-Object System.Net.WebClient;  $client.Proxy.Credentials =[System.Net.CredentialCache]::DefaultNetworkCredentials; $client.DownloadFile($a, $f)}" ""%2"" ""%1""
[...]

PR #2418, opened 2018-04-17
Tell Win7 users to upgrade PowerShell

winks:
Related to #2412 and maybe #2416 - also ref #2405


technomancy:
Looks good; thanks!

PR #2419, opened 2018-04-21
Bump a few more deps before 2.8.2 release

ehashman:
- Pomegranate -> 1.0.0 to use new dynapath properly
- New patch version for slf4j
- Made deps pedantic :+1: 
- Updated the leiningen-core pom.xml
- Updated the bootclasspath-deps.clj


technomancy:
Looks great; thanks!

PR #2431, opened 2018-05-12
Exclude project-specific `profiles.clj` from version control

### LINKED ISSUE #2429: Ignore project-specific `profiles.clj` in version control
From [`PROFILES.md`](https://github.com/technomancy/leiningen/blob/2.8.1/doc/PROFILES.md#declaring-profiles):

> In addition to `project.clj`, profiles also can be specified in `profiles.clj` within the project root. Profiles specified in `profiles.clj` will override profiles in `project.clj`, so this can be used for project-specific overrides that you don't want committed in version control.

So we should probably add `profiles.clj` to `.gitignore` and `.hgignore` be default.

I'd be happy to open a PR if we agree this is correct.

xiongtx:
Fix #2429.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ TAGS
 test_projects/*/target
 pom.xml
 deps.txt
+profiles.clj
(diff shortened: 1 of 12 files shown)


technomancy:
Good catch; thanks!

PR #2432, opened 2018-05-12
Update `lein change` docstring to include set command

danielcompton:
I found it tricky finding the right language to describe this, happy to update if people would prefer something different.


technomancy:
I think with the example you've used it's pretty clear what's going on. Thanks.

PR #2436, opened 2018-05-30
Update clojars cert

tobias:
This updates the bundled clojars cert to be the currently active
certificate, which expires 2020-06-17.


technomancy:
Thanks!

PR #2442, opened 2018-06-21
Fix typo in French irony

ehashman:
I think this should be eaux, like "gateaux" or "châteaux"


technomancy:
I don't speak French, but I'll take your word for it!

PR #2444, opened 2018-06-26
Update nREPL to 0.4.3

### LINKED ISSUE #2434: Migration plans for nREPL 0.4
That's basically a follow up of nrepl/nREPL#1 and was also discussed in nrepl/nREPL#4.

TL;DR; - In verions 0.4 of the "new" nREPL the namespaces will be changed from `clojure.tools.nrepl.*` to `nrepl.*`. 

As this is a different artifact technically speaking(`nrepl/nrepl` instead of `clojure.tools/nrepl`) I think the best migration path for Leiningen would be to just start depending on both projects and have two tasks for the two nREPLs until we can drop the old version (e.g. `lein repl` and `lein nrepl`). The second task can just start a headless server until reply adds support for `nrepl/nrepl`. 

How does this sound?

bbatsov:
As no one really commented on #2434 I've decided to try an alternative approach. Here's my preferred (quick) solution to the migration to the new nREPL. I know this is probably less than ideal, but it's a one time deal, and I'd really appreciate your support as I rather be dealing with nREPL improvements than with migration paths.

nREPL was recently moved out of clojure-contrib and its development
continued under a dedicated org (https://github.com/nrepl). The new project is (and will remain
to be) completely backwards compatible on the nREPL API level, but
the namespace had to be changed.

This fixes #2434.

### DIFF
--- a/resources/leiningen/bootclasspath-deps.clj
+++ b/resources/leiningen/bootclasspath-deps.clj
@@ -54,7 +54,7 @@
  org.apache.maven/maven-resolver-provider "3.5.0"
  org.clojure/data.xml "0.0.8"
  org.clojure/tools.macro "0.1.5"
- nrepl "0.4.1g"
+ nrepl "0.4.1"
  org.codehaus.plexus/plexus-component-annotations "1.7.1"
  org.codehaus.plexus/plexus-interpolation "1.24"
  org.codehaus.plexus/plexus-utils "3.0.24"
--- a/resources/leiningen/bootclasspath-deps.clj
+++ b/resources/leiningen/bootclasspath-deps.clj
@@ -54,7 +54,7 @@
  org.apache.maven/maven-resolver-provider "3.5.0"
  org.clojure/data.xml "0.0.8"
  org.clojure/tools.macro "0.1.5"
- org.clojure/tools.nrepl "0.2.12"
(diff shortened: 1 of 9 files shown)


bbatsov:
See also https://github.com/boot-clj/boot/pull/703

bbatsov:
I've also created one simple plugin, so people can easily use nREPL 0.4, while waiting for this to be merged. (https://github.com/nrepl/lein-nrepl)

resources/leiningen/bootclasspath-deps.clj
```diff
@@ -54,7 +54,7 @@
  org.apache.maven/maven-resolver-provider "3.5.0"
  org.clojure/data.xml "0.0.8"
  org.clojure/tools.macro "0.1.5"
- org.clojure/tools.nrepl "0.2.12"
+ nrepl "0.4.1g"
```
razum2um: @bbatsov a typo

bbatsov: Fixed.

bbatsov:
I've updated the PR to nREPL 0.4.3.

technomancy:
Looks good; thanks.

bbatsov:
🙇

Issue #2433, opened 2018-05-19
Support Clojure 1.9 in plugins.


DLaps:
The current release depends on Clojure 1.8. I think there is no possibility to use new built-in functionality in plugins like clojure.spec.alpha, which is only part of Clojure 1.9.

A new release of leiningen, which supports Clojure 1.9 in plugins, would be nice.

PsukheDelos:
Just putting our interest into this out there as it's making building our application increasingly difficult. Is there any traction on when this might be possible / is it in the pipeline at all? Really love Leiningen by the way, keep up the good work! -G

technomancy:
The likelihood of plugin breakage is too high for this to be included in Leiningen 2.8.2, but I'll merge it once that one is released so it can be part of the following version.

technomancy:
Sure, I'm hoping to get 2.8.2 out in a few weeks; the only open blockers are listed at https://github.com/technomancy/leiningen/milestone/26

Beyond that I don't have any concrete plans, but hopefully you can test your changes on the master branch at that point and we can get some wider testing on it to get a feel for what kind of potential plugin incompatibility it might introduce.

PsukheDelos:
Legend! Thanks so much @technomancy! -G

danielcompton:
There is an open PR for this at https://github.com/technomancy/leiningen/pull/2449, although it doesn't look like it is in a mergeable state.

technomancy:
Done in 9699fd7.

PR #2449, opened 2018-07-18
updated base clojure rev to 1.9 technomancy/leiningen#2433

eoliphan:
Updated clojure version, all tests pass


ehashman:
Two comments:

1. The root project.clj needs to be updated too. It's possible that its 1.8 requirement overriding this one which would make this a no-op. Have you checked `lein deps :tree` with the change?
2. This may break plugins. We should probably make a community announcement and release at least a minor version (rather than patch).
3. Relatedly, I'd prefer to release what we have in master before trying to merge a Clojure 1.9 upgrade.

technomancy:
9699fd7 takes care of this.

Issue #1890, opened 2015-04-29
Obscure "no connector" error when repository not configured correctly.


timmc:
I had the following code in my project.clj -- note the two repositories occupying the same vector:

```
  :deploy-repositories
  [["releases" {:url "https://oss.sonatype.org/service/local/staging/deploy/maven2/"
                :creds :gpg}
    "snapshots" {:url "https://oss.sonatype.org/content/repositories/snapshots/"
                 :creds :gpg}]]
```

When I ran `lein deploy`, instead of a message about the `snapshots` repository not being configured, I get a warning about missing credentials and then an error about there not being a Maven connector:

```
timmc@bc-timmc:~/libre/johnny$ lein deploy
No credentials found for snapshots(did you mean `lein deploy clojars`?)
See `lein help deploying` for how to configure credentials to avoid prompts.
Username: timmc
Password: 
Compiling 49 source files to /home/timmc/libre/johnny/target/classes
Created /home/timmc/libre/johnny/target/johnny-0.1.0-SNAPSHOT.jar
Wrote /home/timmc/libre/johnny/pom.xml
Failed to deploy artifacts/metadata: No connector available to access repository snapshots (snapshots) of type default using the available factories FileRepositoryConnectorFactory, WagonRepositoryConnectorFactory
```

I think we could probably provide a clearer error than that.

marick:
For people finding this page via a search for `"No connector available to access repository snapshots" lein leiningen` who are also inexpert in things Maven-ey: 

This is an error message you can get by deploying a *major-minor-tiny*-SNAPSHOT version to clojars via `lein deploy` when you don't have a `["snapshots" :clojars]` entry in `:deploy-repositories` in your `project.clj`. If you make that latter look like the following, `lein deploy` will work:

```
  :deploy-repositories [["releases" :clojars]
                        ["snapshots" :clojars]]
```

Alternately, you can just use `lein deploy clojars`.

technomancy:
This is unfortunate, but I don't think it's worth handling this particular potential mistake specifically; better to handle it in #2223.

timmc:
Makes sense! And the error is searchable here in the meantime. :-)

PR #2451, opened 2018-07-22
Update DEPLOY.md as proposed in #1890

j-cr:
This is a tiny documentation change that can save newcomers some googling. With the proposed change, the confusing error described in #1890 won't be encountered by those who follow the DEPLOY.md docs.

See https://github.com/technomancy/leiningen/issues/1890#issuecomment-174192434

>marick commented on Jan 23, 2016
> For people finding this page via a search for "No connector available to access repository snapshots" lein leiningen who are also inexpert in things Maven-ey:
> 
> This is an error message you can get by deploying a major-minor-tiny-SNAPSHOT version to clojars via lein deploy when you don't have a ["snapshots" :clojars] entry in :deploy-repositories in your project.clj. If you make that latter look like the following, lein deploy will work:

### DIFF
--- a/doc/DEPLOY.md
+++ b/doc/DEPLOY.md
@@ -236,7 +236,8 @@ order to make `lein deploy` with no argument target Clojars, include this in
 your `project.clj`:
 
 ```clj
-{:deploy-repositories [["releases" :clojars]]}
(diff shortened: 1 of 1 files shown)


technomancy:
Sure; makes sense. thanks.

Issue #2453, opened 2018-07-24
Support network-centric project overrides (:mirrors, :repositories, etc) outside of profiles  


ghaskins:
Certain network-centric features, such as :mirrors have a chicken and egg problem:

- They often can be environment specific (e.g. local-dev vs CI execution), meaning they belong in a profile-like mechanism as opposed to the project file.
- The profiles mechanism needs the network before the profiles are merged to facilitate plugins that modify the profiles themselves.  Therefore, keys such as :mirrors are ignored within profiles until after network operations have already begun.

In discussions with @technomancy on freenode, it was thought that the problem could be addressed by a similar-yet-distinct mechanism from profiles, geared towards (and constrained to) early-stage network configuration.

The proposal is therefore to introduce a simple map with keys for:
- :repositories
- :mirrors
- :certificates
- :plugin-repositories
- :local-repo

These keys can be specified in an EDN file repository-overrides.clj within either ~/.lein/ or $projectdir, with the latter eclipsing the former when both exist.  If an override is found, the keys will be merged into the project early on, before any network operations are performed, including the loading of plugins.  We will select-keys from the above list, and report a WARNING for any keys outside of this set that are being ignored.

technomancy:
This could still use some documentation if you're up for it.

technomancy:
This is working now.

PR #2454, opened 2018-07-24
[LEIN-2453] Support repository-overrides.clj

ghaskins:
This patch adds support for a project override called
repository-overrides.clj.  The details for this feature
are captured here:

https://github.com/technomancy/leiningen/issues/2453

The current implementation only supports projectdir based
overrides, but support for other profile locations such
as ~/.lein can be added in the future.

Signed-off-by: Greg Haskins <gregory.haskins@gmail.com>


technomancy:
Looks good, but could you also add a description of the functionality to `doc/FAQ.md`?

PR #2456, opened 2018-08-05
Upgrade data.xml to new 0.2.0-alpha5

ehashman:
`org.clojure/data.xml 0.2.0-alpha5` has some exciting new features and bugfixes.

It's also the version of org.clojure/data.xml used by tools.deps.alpha. I'd like both in Debian and they need to share deps for head-to-head to work :)

https://github.com/clojure/tools.deps.alpha/blob/744c9c027b1b0d71081a7903b517abbd960a9849/pom.xml#L94-L98

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -9,7 +9,7 @@
   ;; If you update these, update resources/leiningen/bootclasspath-deps.clj too
   :dependencies [[leiningen-core "2.8.2-SNAPSHOT"]
                  ;; needed for pom
-                 [org.clojure/data.xml "0.0.8"]
+                 [org.clojure/data.xml "0.2.0-alpha5"]
                  ;; needed for test
                  [bultitude "0.2.8"]
                  ;; needed for new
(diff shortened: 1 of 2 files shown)


ehashman:
Looks like the API change was breaking... I'll see if I have some time to take a stab to fix that up.

glts:
I could do the changes in `leiningen.pom` if there is interest.

glts:
Doing a bit of spring cleaning here … I discovered an old branch where this upgrade is done. I have updated the branch and am linking it here for your consideration: https://github.com/glts/leiningen/tree/data.xml-0.2.0-alpha5.

Unfortunately, this is a large change, here’s what I noted down when I wrote this (last year):

> I have a bit of experience with the new data.xml, so was motivated to look into this and push a commit.
>
> Between 0.0.8 and 0.2.0-alpha5 data.xml gained XML namespace support and was more or less rewritten by the new maintainer. I think the version bump is welcome, but now all XML should be namespaced. The large diff is due to the qualifying of all XML element literals with the `::pom` namespace.
>
> …
>
> In terms of impact, POM output should be unchanged (except for a newline after the XML declaration). Of course, if someone is using data.xml in their Leiningen plugin, it might break (right?).

Hope you don’t mind, cheers

ehashman:
@glts if you wanna pick this up and run with it, do feel free. I doubt I'll have any time soon to follow up on this.

glts:
@ehashman Actually I am keen, so will open a pull request. Cheers

ehashman:
Thanks for running with this!

PR #2468, opened 2018-10-15
Update bultitude dependency to 0.3.0 & new maintainer

TimoFreiberg:
I started maintaining bultitude recently and fixed an [issue](https://github.com/TimoFreiberg/bultitude/issues/4) described here: https://github.com/TimoFreiberg/bultitude/pull/3

> lein test breaks on files that start with in-ns, and which refer to keyword
> abbreviations (::foo/bar) and namespaced map abbreviations (#::foo{:bar :baz}).

I am not sure about the edit to the `resources/leiningen/bootclasspath-deps.clj` file, when I ran the commented code inside it generated a map with several differences to the literal in the file.

### DIFF
--- a/leiningen-core/project.clj
+++ b/leiningen-core/project.clj
@@ -5,7 +5,7 @@
   :description "Library for core functionality of Leiningen."
   ;; If you update these, update resources/leiningen/bootclasspath-deps.clj too
   :dependencies [[org.clojure/clojure "1.8.0"]
-                 [bultitude "0.2.8" :exclusions [org.tcrawley/dynapath]]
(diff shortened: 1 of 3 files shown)


TimoFreiberg:
> I've tried this out in our codebase, which has unit test namespaces split among many clj files, and I can confirm that abbreviated keywords `::foo/bar` work properly.

Just to make sure I understand you correctly:
Are you confirming that my PR fixes the issue? Or are you saying that you cannot reproduce the problem?

csm:
Sorry I wasn’t clear: this now works with your fixes.

technomancy:
Thanks for submitting it, and for the confirmation that it's working.

PR #2470, opened 2018-10-26
Exclude dot files at any level of the jar hierarchy.

brunchboy:
This could have been done with a single regular expression, but at too high a cost in readability, in my estimation.

This approach does work, and does not break any existing tests. I could not find any tests for the existing `:jar-exclusions` functionality on which to model a new test for this fix, and it would have taken too deep a dive for me to figure out how to write one from scratch, I'm afraid.


technomancy:
Looks good; thanks for submitting this.

PR #2474, opened 2018-11-01
Update reply

bbatsov:
There was a small problem in reply 0.4.1 that's addressed in 0.4.2 (https://github.com/trptcolin/reply/commit/7fe67e4e2eb5580e42da6662fd05cd4c62a6533c). 

This also updates nREPL to its most recent version.


technomancy:
Looks good; thanks.

PR #2481, opened 2018-11-20
Updated Lein.ps1

JamesTryand:
This is such that the DefaultProxy is no longer working for Powershell core 6, so added workaround for these versions. Verified to be working as expected now.


technomancy:
I see. Looks good as the old behavior should still be used in cases where it works. Thanks.

PR #2493, opened 2018-12-01
Allow freeform :pom-plugin configuration

### LINKED ISSUE #2489: :pom-plugins allows only limited plugin configuration
I'm trying to achieve the following XML structure in my pom.xml:

```
      <plugin>
        <groupId>com.google.appengine</groupId>
        <artifactId>appengine-maven-plugin</artifactId>
        <version>1.9.68</version>
        <configuration>
          <project>foo</project>
          <version>bar</version>
        </configuration>
      </plugin>
```

I'd expect this config to work:

```
  :pom-plugins [[com.google.appengine/appengine-maven-plugin "1.9.68"
                 [:configuration
                  [:project "foo"]
                  [:version "bar"]]]])
```

But alas, the third parameter isn't included as is but mangled such that having two second-level children is impossible (AFAICT). See https://github.com/technomancy/leiningen/blob/master/src/leiningen/pom.clj#L255

I'd be glad to post a pull request that either

- includes vectors as-is (as I would have expected)
- includes lists as-is (doesn't break any existing configs and provides a workaround)
- adds a fourth parameter that is included as-is

Or something else, if none of these is applicable.

dancek:
Lists are passed through without alteration. Fixes #2489.

Note that in current master, there's one failing test. Other tests should pass, including the ones I added.

Any feedback is welcome.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,7 +11,8 @@
 * Add pprint `--not-pretty` argument that prints instead of pprinting. (Rob Browning)
 * Always send diagnostic messages to standard error. (Rob Browning)
 * Add project coordinate data to jar metadata. (Conor McDermottroe)
-S
+* Allow freeform `:pom-plugin` configuration. (Hannu Hartikainen)
+
 ## 2.8.1 / 2017-10-27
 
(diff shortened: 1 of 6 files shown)


technomancy:
Thanks! I missed the test failure because the IRC notification is broken, appreciate the heads-up.

PR #2496, opened 2018-12-07
[Fix #2492] Respect nREPL's configuration files

### LINKED ISSUE #2492: Respect nREPL configuration files
nREPL introduced global and project-local config files in version 0.5. The details are here https://nrepl.xyz/nrepl/usage/server.html#_server_configuration

Ideally `lein repl` should honour them when it's spinning up a new server. I think they should be treated more or less the same way as the existing `repl-options` (which were the inspiration for this feature) - values supplied via `repl-options` should have the highest precedence, but if they are not present we can fallback to whatever is in the config files. The configuration is exposed by `nrepl.config` as a simple map.

bbatsov:
This change makes leiningen consult nREPL config files, in addition to
its nREPL config mechanisms (env variables and the :repl-options).

The changes are backwards compatible and any configuration coming from
Lein's own mechanisms in going to take precedence over something
specified via nREPL's config files.

In addition to this, we're also making it easy to specify the
transport to use with nREPL. Extremely useful when you have a client
that doesn't support bencode, but works fine with some of the other
available transports.

### DIFF
--- a/leiningen-core/src/leiningen/core/project.clj
+++ b/leiningen-core/src/leiningen/core/project.clj
@@ -554,7 +554,7 @@
                 :test-selectors {:default (with-meta '(constantly true)
                                             {:displace true})}
                 ;; bump deps in leiningen's own project.clj with these
-                :dependencies '[^:displace [nrepl/nrepl "0.5.1"
+                :dependencies '[^:displace [nrepl/nrepl "0.5.2"
                                             :exclusions [org.clojure/clojure]]
                                 ^:displace [clojure-complete "0.2.5"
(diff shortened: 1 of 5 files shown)


technomancy:
Looks good; thanks. One minor redundancy with `leiningen.core.utils/require-resolve` but I can take care of that. Thanks!

bbatsov:
> Looks good; thanks. One minor redundancy with leiningen.core.utils/require-resolve but I can take care of that. Thanks!

Seems every project has one of those, but now Clojure 1.10 is finally adding this upstream.  :-)

PR #2509, opened 2018-12-16
[leiningen-core] Removed explicit deps and checked dependencies

cnuernber:
Simply removed explicitly set transitive dependencies allowing more things to be up to date.

Mainly testing out [clj-commons/vizdeps](https://github.com/clj-commons/vizdeps/tree/upgrade-clojure-1.9.0).

Tests pass.

[before-dependencies.pdf](https://github.com/technomancy/leiningen/files/2683842/before-dependencies.pdf)

[after-implicit.pdf](https://github.com/technomancy/leiningen/files/2683843/after-implicit.pdf)

### DIFF
--- a/leiningen-core/project.clj
+++ b/leiningen-core/project.clj
@@ -8,12 +8,7 @@
                  [timofreiberg/bultitude "0.3.0" :exclusions [org.tcrawley/dynapath]]
                  [org.flatland/classlojure "0.7.1"]
                  [robert/hooke "1.3.0"]
-                 [com.cemerick/pomegranate "1.1.0"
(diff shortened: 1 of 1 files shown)


technomancy:
@ehashman do you remember if there was a specific reason we were sticking with 2.x of `wagon-http`? If this is going to create a bunch of annoying packaging work I don't want to pull in something new unless there's some actual bugfix we would benefit from.

ehashman:
2.12 had a security fix that pomegranate's wagon lacked. But now pomegranate should be on wagon 3.x.

technomancy:
Cool; this should be fine then. Thanks.

PR #2510, opened 2018-12-17
Fix repl initialization with a different transport

pfeodrippe:
If the transport is set and it's not part of nREPL itself,
it has to be required before starting the server.

### DIFF
--- a/src/leiningen/repl.clj
+++ b/src/leiningen/repl.clj
@@ -199,27 +199,29 @@
        (catch Throwable e#))))
 
 (defn- server-forms [project cfg ack-port start-msg?]
-  [`(let [server# (nrepl.server/start-server
-                   :bind ~(:host cfg)
-                   :port ~(:port cfg)
-                   :transport-fn ~(:transport cfg)
-                   :greeting-fn ~(:greeting-fn cfg)
-                   :ack-port ~ack-port
-                   :handler ~(handler-for project))
-          port# (:port server#)
-          repl-port-file# (apply io/file ~(repl-port-file-vector project))
-          ;; TODO 3.0: remove legacy repl port support.
-          legacy-repl-port# (if (.exists (io/file ~(:target-path project "")))
-                              (io/file ~(:target-path project) "repl-port"))]
-      (when ~start-msg?
-        (println "nREPL server started on port" port# "on host" ~(:host cfg)
-                 (str "- "
-                      (nrepl.transport/uri-scheme ~(or (:transport cfg) #'nrepl.transport/bencode))
-                      "://" ~(:host cfg) ":" port#)))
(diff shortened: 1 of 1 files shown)


pfeodrippe:
You will be able to use `lein repl :headless` with new transports.

`lein repl` calls `REPL-y` which does is not updated yet.

bbatsov:
We already require and resolve the transport var here https://github.com/technomancy/leiningen/blob/master/src/leiningen/repl.clj#L44

I assume this is enough, right?

src/leiningen/repl.clj
```diff
@@ -199,27 +199,29 @@
...
-      (when ~start-msg?
-        (println "nREPL server started on port" port# "on host" ~(:host cfg)
-                 (str "- "
-                      (nrepl.transport/uri-scheme ~(or (:transport cfg) #'nrepl.transport/bencode))
-                      "://" ~(:host cfg) ":" port#)))
-      (spit (doto repl-port-file# .deleteOnExit) port#)
-      (when legacy-repl-port#
-        (spit (doto legacy-repl-port# .deleteOnExit) port#))
-      @(promise))
+  [`(do (if ~(some-> (:transport cfg) meta :ns str)
```
bbatsov: Keep in mind this is not nREPL's raw config, but something that has already been processed it went to Lein's `cfg` map.

src/leiningen/repl.clj
```diff
@@ -252,11 +254,17 @@
      `(do (try (require '~(init-ns project)) (catch Exception t#))
           (require ~@(init-requires project 'reply.main))))))
 
-(def ack-server
+(defn ack-server
```
bbatsov: Why do we care about the transport of the ack server at all? It simply has to acknowledge one message - I guess assuming it's using bencode is perfectly fine.

src/leiningen/repl.clj
```diff
@@ -252,11 +254,17 @@
...
   "The server which handles ack replies."
-  (delay (nrepl.server/start-server
-          :bind (repl-host nil)
-          :handler (nrepl.ack/handle-ack nrepl.server/unknown-op))))
+  [transport]
+  (if transport
```
bbatsov: Assuming we care about the transport - I think that passing `nil` will select `bencode`, which would simplify this code.

src/leiningen/repl.clj
bbatsov: Ah, yeah. I keep forgetting how Lein runs the plugins in a different process. I guess you should add some explanations for this for future reference. I'd also try to reduce this code duplication in the `some->` and the `require`.

bbatsov:
Apart from my minor remark about the code duplication everything looks good. I guess you should also mention this in the changelog.

technomancy:
Thanks for looking at this.

PR #2515, opened 2018-12-21
[Fix #2514] Mention the switch to nREPL 0.5 in the changelog

### LINKED ISSUE #2514: Mention switching to nrepl/nrepl in 2.8.2+ changelog
This could be breaking change for nREPL middleware and some plugins which rely on nREPL stuff.

bbatsov:


technomancy:
Cool; thanks.

PR #2525, opened 2019-01-20
Add `scheme` at reply options and transport to ack-server

pfeodrippe:
# Have to wait for the below referenced PR so this can be merged

Repl-y does not accept a scheme option at actual version,
but a PR there (https://github.com/trptcolin/reply/pull/194)
is being developed for it.

The changes here add support to pass the `scheme` to repl-y
so `lein repl` can start without problems for new transports
(see more at https://github.com/nrepl/fastlane).

### DIFF
--- a/src/leiningen/repl.clj
+++ b/src/leiningen/repl.clj
@@ -7,7 +7,7 @@
             nrepl.ack
             nrepl.config
             nrepl.server
-            nrepl.transport
+            [nrepl.transport :as transport]
             [cemerick.pomegranate :as pomegranate]
             [leiningen.core.eval :as eval]
             [leiningen.core.main :as main]
@@ -118,7 +118,7 @@
               (s/join ":" x)
               (ensure-port x))))))
 
-(defn options-for-reply [project & {:keys [attach port]}]
+(defn options-for-reply [project & {:keys [attach port scheme]}]
   (as-> (:repl-options project) opts
         (merge {:history-file (->> (if-let [root (:root project)]
                                      [root ".lein-repl-history"]
@@ -136,7 +136,8 @@
                           :else {}))
         (clojure.set/rename-keys opts {:prompt :custom-prompt
                                        :welcome :custom-help})
-        (if (:port opts) (update-in opts [:port] str) opts)))
+        (if (:port opts) (update-in opts [:port] str) opts)
+        (if scheme (assoc opts :scheme scheme) opts)))
 
 (defn init-ns [{{:keys [init-ns]} :repl-options, :keys [main]}]
   (or init-ns (if main (if (namespace main)
@@ -198,6 +199,10 @@
          (proxy [sun.misc.SignalHandler] [] (handle [signal#])))
        (catch Throwable e#))))
 
+(defn- cfg->transport-uri-scheme
+  [cfg]
+  (transport/uri-scheme (or (:transport cfg) #'transport/bencode)))
+
 (defn- server-forms [project cfg ack-port start-msg?]
   [`(do (if ~(some-> (:transport cfg) meta :ns str)
           (require (symbol ~(-> (:transport cfg) meta :ns str))))
@@ -216,7 +221,7 @@
           (when ~start-msg?
             (println "nREPL server started on port" port# "on host" ~(:host cfg)
                      (str "- "
-                          (nrepl.transport/uri-scheme ~(or (:transport cfg) #'nrepl.transport/bencode))
+                          (transport/uri-scheme ~(or (:transport cfg) #'transport/bencode))
                           "://" ~(:host cfg) ":" port#)))
           (spit (doto repl-port-file# .deleteOnExit) port#)
           (when legacy-repl-port#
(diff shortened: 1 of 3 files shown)


pfeodrippe:
Will reopen it when the other PR is merged at REPL-y

pfeodrippe:
I'll test this one again, reopening it o/

bbatsov:
@pfeodrippe Seems we completely forgot about this. Can you look into the test failures and add some changelog entry?

pfeodrippe:
I'll test it today and will report if find any problem o/, thanks for remembering me

pfeodrippe:
@bbatsov Done with testing, was missing a branch in a `if` 🤦 

You may have more context than me, do you recommend some not so trivial manual testing?

src/leiningen/repl.clj
```diff
@@ -314,7 +321,18 @@
                                 :repositories (map classpath/add-repo-auth
                                                    (:repositories project)))
   (let [launch (utils/require-resolve 'reply.main/launch-nrepl)]
-    (launch (options-for-reply project :attach attach)))  )
+    (launch (options-for-reply project :attach attach))))
+
+(defn client-with-cfg [project cfg attach]
```
bbatsov, line 326: @pfeodrippe I'm a bit confused by this part. Why is all this needed?

bbatsov:
> You may have more context than me, do you recommend some not so trivial manual testing?

I guess now that the EDN transport is in you can add some test that spins it. I'm guessing this can also be done in reply. But if @technomancy is fine with only manual testing that's fine by me.

src/leiningen/repl.clj
bbatsov: Ah, I see you've modified the existing `client` function. Isn't it easier to just one more arity for it instead of copying it?

bbatsov: I actually doubt it has usages outside of Lein, so probably it can be modified directly, although the 2 arity version serves fine our purposes.

src/leiningen/repl.clj
```diff
@@ -264,11 +269,13 @@
      `(do (try (require '~(init-ns project)) (catch Exception t#))
           (require ~@(init-requires project 'reply.main))))))
 
-(def ack-server
+(defn- ack-server
```
technomancy, line 272: If you're changing this to a function instead of a top-level def, you should remove the `delay` because that was only added because it wasn't a function.

src/leiningen/repl.clj
```diff
@@ -198,6 +199,10 @@
          (proxy [sun.misc.SignalHandler] [] (handle [signal#])))
        (catch Throwable e#))))
 
+(defn- cfg->transport-uri-scheme
+  [cfg]
+  (nrepl.transport/uri-scheme (or (:transport cfg) #'nrepl.transport/bencode)))
```
technomancy, line 204: Please add an `:as` alias for this if we're going to use its functions in this file; thanks.

src/leiningen/repl.clj
technomancy: Yeah, this would be better as an additional arity to the existing `client` function since it mostly does the same thing; thanks.

pfeodrippe:
@bbatsov @technomancy Review fixes applied, thank you!

About the test, is there any namespace where I can add a shell or `eval` test? I've tried to look at the project, but couldn't find.

I could add this "command line" test ns if there isn't  o/

bbatsov:
> About the test, is there any namespace where I can add a shell or eval test? I've tried to look at the project, but couldn't find.

I didn't find anything as well. I did notice, however, that in the repl unit test namespace we didn't add anything about the transport, so I guess you can add a few related tests there. 

I'm thinking that REPL-y's test suite might be a better place for an integration test, as it already has some of those.

src/leiningen/repl.clj
```diff
@@ -302,19 +309,28 @@
...
 
-(defn client [project attach]
-  (when (is-uri? attach)
-    (require 'drawbridge.client))
-  (pomegranate/add-dependencies :coordinates (:dependencies reply-profile)
-                                :repositories (map classpath/add-repo-auth
-                                                   (:repositories project)))
-  (let [launch (utils/require-resolve 'reply.main/launch-nrepl)]
-    (launch (options-for-reply project :attach attach)))  )
+(defn resolve-reply-launch-nrepl
```
pfeodrippe, line 317: Created to facilitate tests

pfeodrippe:
> I didn't find anything as well. I did notice, however, that in the repl unit test namespace we didn't add anything about the transport, so I guess you can add a few related tests there.
I'm thinking that REPL-y's test suite might be a better place for an integration test, as it already has some of those.

Great, I'll add try to add integration tests there soon then o/

Added some mocked test here

bbatsov:
> Great, I'll add try to add integration tests there soon then o/

Excellent!

> Added some mocked test here

Looks good to me. To me it seems the PR is in a good shape to be merged.

technomancy:
Thanks; looks good!

PR #2528, opened 2019-02-08
Bump nREPL to 0.6.0

ehashman:
This drops the transitive bencode dependency (and ensures clj 1.10 support in nREPL).


technomancy:
Nice; thanks!

PR #2532, opened 2019-02-14
 Use absolute LEIN_DIR path when running from checkout

juergenhoetzel:
Fixes invalid relative path resolution to `lein` command from checkout:
```
juergen@samson:~/clojure → LANG=C ./leiningen/bin/lein
Recalculating Leiningen's classpath.
./leiningen/bin/lein: line 197: ./leiningen/bin/lein: No such file or directory
sum: ./leiningen/project.clj: No such file or directory
sum: ./leiningen/leiningen-core/project.clj: No such file or directory
```


technomancy:
Makes sense; thanks.

PR #2534, opened 2019-02-15
Add GitHub issue templates

danielcompton:
Many of the bugs that come up could be more easily triaged if the reporter provided better context for the issue that they were facing. This PR adds issue templates for bugs, features, and miscellaneous issues.

Happy to change the wording in the bug report or add/remove sections. 

You can see an example of what this looks like when reporting a bug by going to https://github.com/clojars/clojars-web/issues/new/choose.


technomancy:
Good idea; thanks.

PR #2535, opened 2019-02-15
rm "master" from :repl-options :prompt example

apetro:
Avoid avoidable nod to "master-slave" idiom in sample `project.clj` configuration.

Resulting shortened example

+ demonstrates the same syntax, string concatenation wrapping around confirmation of the `ns`
+ is still relatable. One can still read the example and understand it to be a function that takes `ns` and returns a string and that string is a repl prompt.
+ is shorter. hurray for brevity
+ avoids an avoidable "master" mention, as in "master-slave", as in disturbing.


technomancy:
The intent here was to evoke more of a butler scenario, but I see how it could be misunderstood.

PR #2536, opened 2019-02-17
Handle empty http(s)_proxy string

### LINKED ISSUE #2426: http_proxy unset vs empty string
I ran into an issue with `http_proxy` being set to an empty string("") when getting dependencies.
For example if I use `lein deps` I get the following...

```bash
> lein deps
Could not transfer artifact clj-time:clj-time:jar:0.11.0 from/to central (https://repo1.maven.org/maven2/): Connect to  [localhost/127.0.0.1] failed: Connection refused (Connection refused)
Could not transfer artifact clj-time:clj-time:jar:0.11.0 from/to clojars (https://repo.clojars.org/): Connect to  [localhost/127.0.0.1] failed: Connection refused (Connection refused)
```
The solution is just to unset `http_proxy`. This sort of caught me off guard because it seems like other tools consider an empty string in `http_proxy` as ignorable. For example [curl](https://curl.haxx.se/libcurl/c/CURLOPT_PROXY.html).

> Setting the proxy string to "" (an empty string) will explicitly disable the use of a proxy, even if there is an environment variable set for it.

A few others follow this pattern like `apt` and here is an [issue with rubygems](https://github.com/rubygems/rubygems/issues/1706) discussing a similar topic.

The suggestion would be to ignore `http_proxy` when it is an empty string.

If is this something the maintainers of lein would be ok with changing, I am willing to create the pull request. In any case I figured someone else might run into this issue and I wanted to document it somewhere.

Leiningen 2.8.1 on Java 1.8.0_151 OpenJDK 64-Bit Server VM
Linux Mint 18.1

juergenhoetzel:
Fixes #2426.

### DIFF
--- a/leiningen-core/src/leiningen/core/classpath.clj
+++ b/leiningen-core/src/leiningen/core/classpath.clj
@@ -233,7 +233,8 @@
   "Returns a map of the JVM proxy settings"
   ([] (get-proxy-settings "http_proxy"))
   ([key]
-     (if-let [proxy (System/getenv key)]
+   (let [proxy (System/getenv key)]
+     (when-not (str/blank? proxy)
        (let [url (utils/build-url proxy)
              user-info (.getUserInfo url)
              [username password] (and user-info (.split user-info ":"))]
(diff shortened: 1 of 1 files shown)


technomancy:
Looks good; thanks.

Issue #2521, opened 2019-01-04
aot and main: main compiled twice


pesterhazy:
Setting aot and main leads to unexpected behavior.

Actual behavior:

In a [sample project](https://github.com/pesterhazy/lein-aot-example/blob/7314770e1ac941276ec8645af269598afd281997/project.clj#L6) where I'm setting `:aot [foo.main]` and `:main foo.main`, the foo.main ns is compiled twice:

```
$ lein uberjar
Compiling foo.main
ns-init foo.main
Compiling foo.main
ns-init foo.main
Created /Users/user/prg/lein-aot-example/target/foo-0.1.0-SNAPSHOT.jar
Created /Users/user/prg/lein-aot-example/target/foo-0.1.0-SNAPSHOT-standalone.jar
```

Expected behavior:

The ns is only compiled once (and the ns-level forms are run only once).

This doesn't happen when I set `:aot :all` but in that case lein compiles all the clj files in the classpath. I only want to AOT one namespace, foo.main.

pesterhazy:
I discovered that I can prevent the ns from being compiled twice by specifying

`:main ^:skip-aot foo.main`

technomancy:
Yeah looks like a bug. Happy to take a patch to make `:main` remove the entry from `:aot` if it's there.

pesterhazy:
@technomancy added a a PR

PR #2541, opened 2019-02-27
Add main to aot nss only if it's not there already

pesterhazy:
Fixes https://github.com/technomancy/leiningen/issues/2521


technomancy:
Looks good; thanks.

PR #2544, opened 2019-03-19
Actually check the checksum of downloaded jar

Vaelatern:
This is a 6 year old TODO comment, it's about time to actually check the checksum.

The checksum as provided is for 2.9.1, because I wanted it for my own use.

### DIFF
--- a/bin/lein
+++ b/bin/lein
@@ -362,15 +362,7 @@ else
     if [ "$LEIN_FAST_TRAMPOLINE" != "" ] && [ -r project.clj ]; then
         INPUTS="$* $(cat project.clj) $LEIN_VERSION $(test -f "$LEIN_HOME/profiles.clj" && cat "$LEIN_HOME/profiles.clj")"
 
-        if command -v shasum >/dev/null 2>&1; then
-            SUM="shasum"
-        elif command -v sha1sum >/dev/null 2>&1; then
-            SUM="sha1sum"
-        else
-            command_not_found "sha1sum or shasum"
-        fi
-
-        INPUT_CHECKSUM=$(echo "$INPUTS" | $SUM | cut -f 1 -d " ")
+        INPUT_CHECKSUM=$(echo "$INPUTS" | $SHASUM_CMD | cut -f 1 -d " ")
         # Just don't change :target-path in project.clj, mkay?
         TRAMPOLINE_FILE="target/trampolines/$INPUT_CHECKSUM"
     else
--- a/bin/lein
+++ b/bin/lein
@@ -9,6 +9,8 @@ function msg {
 }
 
 export LEIN_VERSION="2.9.1"
+# Must be sha256sum, will be replaced by bin/release
+export LEIN_CHECKSUM='ea7c831a4f5c38b6fc3926c6ad32d1d4b9b91bf830a715ecff5a70a18bda55f8'
 
 case $LEIN_VERSION in
     *SNAPSHOT) SNAPSHOT="YES" ;;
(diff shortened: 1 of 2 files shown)


bin/lein
```diff
@@ -75,6 +77,13 @@ function download_failed_message {
 	EOS
 }
 
+function checksum_failed_message {
+    cat <<-EOS 1>&2
+	Failed to properly download $1
+	The checksum was mismatched. and we could not verify the
```
danielcompton: Looks like this sentence wasn't finished?

bin/lein
```diff
@@ -75,6 +77,13 @@ function download_failed_message {
...
+	EOS
```
danielcompton: It could be good to show the expected checksum and the actual checksum one line after the other and aligned, for easy comparison.

bin/lein
```diff
@@ -89,8 +98,15 @@ function self_install {
   $HTTP_CLIENT "$LEIN_JAR.pending" "$LEIN_URL"
   local exit_code=$?
   if [ $exit_code == 0 ]; then
-      # TODO: checksum
-      mv -f "$LEIN_JAR.pending" "$LEIN_JAR"
+      printf "$LEIN_CHECKSUM $LEIN_JAR.pending\n" >> "$LEIN_JAR.pending.shasum"
+      sha256sum -c "$LEIN_JAR.pending.shasum"
```
danielcompton: On macOS `sha256sum` doesn't seem to exist. I can calculate sha256 checksums with `shasum --algorithm 256` though. Is there a portable shasum command for Linux and macOS?

technomancy:
Unfortunately this also breaks the `bin/release` script; unless the checksum is updated as part of the release process (after the uberjar is generated but before the tag is signed) then we will push out new versions of the script which will fail on this error.

bin/lein
Vaelatern: Thanks!

bin/lein
Vaelatern: Sadly, while `shasum` does exist on some Linux systems, others like Debian do not have it.
We could perform a test before this, to choose a sha256sum program.

Vaelatern:
How would you like me to generate these checksums for bin/release, to then copy to this script? We can do some sed -i to insert the shasum in this file.

bin/lein
Vaelatern: Added.

bin/lein
Vaelatern: I reformed to have tests, preferring sha256sum but falling back to shasum --algorithm 256.

Vaelatern:
Poke -- I think all concerns have been addressed, and the code is still minimal.

Vaelatern:
Actually, I don't know how to handle the `bin/release` flow. Is https://github.com/technomancy/leiningen/blob/master/bin/release#L34 the only time `bin/lein` is called, and if so, should the checksum just be updated after that uberjar is created (on line 34, so checksum after https://github.com/technomancy/leiningen/blob/master/bin/release#L36)?

technomancy:
Sorry for the delay on this; I was preparing for a talk at a conference and was heads-down for a while.

> Is L34 the only time bin/lein is called, and if so, should the checksum just be updated after that uberjar is created (on line 34, so checksum after L36)?

Yes, that's correct. It should be a matter of updating the checksum in `bin/lein` right after the uberjar is created in `bin/release`.

One other thing I noticed is that we already have code for detecting whether `shasum` is installed or not on line 335 of `bin/lein`; maybe these two places could be collapsed. Right now the existing code is using sha1 but there is no problem with changing it to use sha256.

Vaelatern:
@technomancy Got an update, I think this is all we need to do to A: update the checksum, and B: only use one type of checksum through `bin/lein`.

technomancy:
Looks good; thank you!

Issue #2549, opened 2019-03-26
:init-ns user leads to stack trace on repl startup


noisesmith:
If init-ns is user, and there is no user.clj on classpath, there is a stacktrace on repl startup.

**To Reproduce**
Using leiningen .2.9.1, create a new project, start a repl and verify it has no errors. Amend the project.clj to have the following contents:
```
(defproject repro "0.1.0-SNAPSHOT"
  :description "FIXME: write description"
  :url "http://example.com/FIXME"
  :license {:name "EPL-2.0 OR GPL-2.0-or-later WITH Classpath-exception-2.0"
            :url "https://www.eclipse.org/legal/epl-2.0/"}
  :dependencies [[org.clojure/clojure "1.10.0"]]
  :repl-options {:init-ns user})
```

**Actual behavior**
The repl errors and shows a stack trace, despite the user namespace being found and used on startup.

**Expected behavior**
I expect to be able to explicitly start up in the user ns, even if I don't have a user.clj / user.cljc

**Logs**
```
$ lein repl
#error {
 :cause Could not locate user__init.class, user.clj or user.cljc on classpath.
 :via
 [{:type java.io.FileNotFoundException
   :message Could not locate user__init.class, user.clj or user.cljc on classpath.
   :at [clojure.lang.RT load RT.java 466]}]
 :trace
 [[clojure.lang.RT load RT.java 466]
  [clojure.lang.RT load RT.java 428]
  [clojure.core$load$fn__6824 invoke core.clj 6126]
  [clojure.core$load invokeStatic core.clj 6125]
  [clojure.core$load doInvoke core.clj 6109]
  [clojure.lang.RestFn invoke RestFn.java 408]
  [clojure.core$load_one invokeStatic core.clj 5908]
  [clojure.core$load_one invoke core.clj 5903]
```

**Environment**
- Leiningen Version: 2.9.1
- Leiningen installation method: manual
- JDK Version: 1.8.0_171-b11

danielcompton:
This also happens in older versions of Leiningen (2.8.1), and older versions of Clojure (1.8.0). @noisesmith did this ever work for you, or is this a brand new issue?

noisesmith:
I discovered this behavior when checking out a project at work. I assume my coworkers all worked around this by defining a `user.clj` file or ignoring the stack trace.

It seems like exempting user.clj from the load logic for init-ns would be a safe option (if it's found, we can rely on clojure loading it, if it's not found, it shouldn't be an error).

[...]

PR #2550, opened 2019-03-27
elides printing of exception when require of user ns throws

noisesmith:
this is one approach for fixing https://github.com/technomancy/leiningen/issues/2549

when require of user fails, it's appropriate to silently ignore that, as user is an implicit ns

~I'm concerned about the step of comparing as strings. The equality check is negative when comparing as symbols, and I'd like to know why.~

### DIFF
--- a/src/leiningen/repl.clj
+++ b/src/leiningen/repl.clj
@@ -225,7 +225,10 @@
    ;; TODO: remove in favour of :injections in the :repl profile
    `(do ~(when-let [init-ns (init-ns project)]
            `(try (doto '~init-ns require in-ns)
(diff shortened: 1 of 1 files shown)


amalloy:
> The equality check is negative when comparing as symbols, and I'd like to know why.

You probably wrote 

    (= '~init-ns 'user)

In a syntax-quoting context, that resolves to

    (= 'user 'leiningen.whatever/user)

at which point it's clear why it's false. Instead, write:

    (= '~init-ns '~'user)

technomancy:
Makes sense; thanks.

PR #2551, opened 2019-03-29
Pick up fix for CVE-2018-10237

nwjsmith:
com.cemerick/pomegranate 1.1.0 includes com.google.guava 20.0 as a
transitive dependency. This is vulnerable to CVE-2018-10237, so we
override it to pick up the fix found in later versions.

### DIFF
--- a/leiningen-core/project.clj
+++ b/leiningen-core/project.clj
@@ -9,6 +9,7 @@
                   :exclusions [org.clojure/clojure]]
                  [org.flatland/classlojure "0.7.1"]
                  [robert/hooke "1.3.0"]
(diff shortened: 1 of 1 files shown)


nwjsmith:
I've opened a PR in pomegranate too, where the root of the problem lies: https://github.com/cemerick/pomegranate/pull/112

technomancy:
Thanks for the heads up on this!

glts:
@nwjsmith This change is incomplete. Notice the comment above your change: ‘If you update these, update resources/leiningen/bootclasspath-deps.clj too’.

Could you update boot classpath deps, too? Cheers

Issue #2554, opened 2019-04-04
What does this sentence mean in section "Checkout Dependencies" of doc/TUTORIAL?


kaznak:
Hello,

I'm now translating doc/TUTORIAL.md to [Japanese](https://github.com/kaznak/leiningen/blob/translate-tutorial-ja/doc/ja/TUTORIAL_ja.md).

While reading it, I did not understand a section titled ["Checkout Dependencies"](https://github.com/technomancy/leiningen/blob/master/doc/TUTORIAL.md#checkout-dependencies).
What does the sentence that the section starts from mean?
That is "Sometimes it is necessary to develop two projects in parallel".

I guess it says that someone develops two or more projects simultaneously, the main one and its dependencies.
And this section says Leiningen can handle dependencies at a time with checkout directory.
Is this correct?

Thank you.

technomancy:
> I guess it says that someone develops two or more projects simultaneously, the main one and its dependencies.
> And this section says Leiningen can handle dependencies at a time with checkout directory.
> Is this correct?

Yes, it means that developing simultaneously can be done without
requiring the tedious install process after every change is made.

kaznak:
Thank you for answering. Now things are cleared.

It was a bit confusable for the first time I read.
This may be the same for some other people,
especially for no native people like me.
So I wrote a patch. I will send it as a pull request.

Thank you again.

PR #2555, opened 2019-04-05
add a detailed explanation into doc/TUTORIAL.md

kaznak:
I wrote a small patch for #2554 .

The main change is as follows.

original:
> Sometimes it is necessary to develop two projects in parallel but ...

patch:
> Sometimes it is necessary to develop two or more projects in parallel,
> the main project and its dependencies, but ...

This patch also contains applying auto line fold on the paragraph.

I think this patch is just a draft and there could be better writing.
If so, please adopt the better one.


technomancy:
Looks good; thanks!

PR #2559, opened 2019-04-17
Tutorial fix

kaznak:
Hi

I may found two mistakes in docs/TUTORIAL.md while translating it.

I'm not a native of English, so don't have absolute confidence.
But if these are mistakes, please merge this patch.

Thank you.


technomancy:
Looks good; thanks.

PR #2569, opened 2019-05-09
Upgrade data.xml to 0.2.0-alpha5

glts:
This change upgrades the data.xml dependency from version 0.0.8 (2014-08-11) to 0.2.0-alpha5 (2017-11-14).

There is no pressing need for doing this upgrade, but still I consider it worthwhile for the benefit of Leiningen plugins, and also justifiable given the recent version bump activity of significant dependencies such as nREPL and Clojure itself. Originally this work was begun by @ehashman in #2456.

Between 0.0.8 and 0.2.0-alpha5, data.xml gained namespace support, a critical feature of any serious XML handling library. The library was basically rewritten by the new maintainer, and then stabilised. For a (subjectively) long time now, 0.2.0-alpha5 has been a solid release of data.xml that I have been very happy to use.

I suggest going with 0.2.0-alpha5 rather than 0.2.0-alpha6, to align with tools.deps.alpha (for the sake of possible Linux dist packaging work, #2456), and because the latter only contains some fixes related to ClojureScript.

Let me know what you think!

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -9,7 +9,7 @@
   ;; If you update these, update resources/leiningen/bootclasspath-deps.clj too
   :dependencies [[leiningen-core "2.9.1"]
                  ;; needed for pom
-                 [org.clojure/data.xml "0.0.8"]
+                 [org.clojure/data.xml "0.2.0-alpha5"]
                  ;; needed for test
                  [timofreiberg/bultitude "0.3.0"
                   :exclusions [org.clojure/clojure]]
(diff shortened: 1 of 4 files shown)


glts:
I scripted a diff between the old and new POMs on a hundred or so random projects from clojure-toolbox.com. The only difference is new newline after the XML declaration (good). Otherwise no diffs, POMs are identical before and after. So this is looking good.

technomancy:
Looks good; thanks for doing that extra work to check compatibility.

glts:
Nice, thank you, and apologies for my hyperactivity here .

Do you think this is worth a release notes entry for its potential impact on Leiningen plugins?

PR #2573, opened 2019-05-13
Add data.codec to bootstrap classpath dependencies after data.xml upgrade

glts:
This adds the dependency `[org.clojure/data.codec "0.1.0"]` to `resources/leiningen/bootclasspath-deps.clj`. This is a new transitive dependency that snuck in with the upgrade of data.xml in #2569.

My apologies for not noticing this earlier! data.codec is only actually used in one unusual code path in data.xml when a byte array is serialised to an XML string.


technomancy:
Sounds good; thanks.

PR #2575, opened 2019-05-15
Include repository policies in POM

### LINKED ISSUE #2377: pom task doesn't include updatePolicy or checksumPolicy
If not, is there a reason it doesn't, or is this just a feature that hasn't yet been implemented?

e.g. 
```
  <repositories>
    <repository>
      <releases>
        <enabled>false</enabled>
        <updatePolicy>always</updatePolicy>
        <checksumPolicy>warn</checksumPolicy>
      </releases>
      <snapshots>
        <enabled>true</enabled>
        <updatePolicy>never</updatePolicy>
        <checksumPolicy>fail</checksumPolicy>
      </snapshots>
      <id>codehausSnapshots</id>
      <name>Codehaus Snapshots</name>
      <url>http://snapshots.maven.codehaus.org/maven2</url>
      <layout>default</layout>
    </repository>
```

glts:
I continue my ramble through the backlog …

This teaches the `pom` task to include update and checksum policies in the generated POM, and so fixes #2377.

### DIFF
--- a/src/leiningen/pom.clj
+++ b/src/leiningen/pom.clj
@@ -201,22 +201,25 @@
             :exclusions exclusions
             :scope scope})]))
 
+(defn- policy-tags [type opts]
+  (seq (keep (partial apply xml-tags)
+             {:enabled (str (if (nil? (type opts))
+                              true
+                              (boolean
+                               (type opts))))
+              :update-policy (or (some-> opts type :update name)
+                                 (some-> opts :update name))
+              :checksum-policy (or (some-> opts type :checksum name)
+                                   (some-> opts :checksum name))})))
+
 (defmethod xml-tags ::repository
   ([_ [id opts]]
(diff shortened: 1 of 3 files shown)


technomancy:
Looks like that should do the trick; thanks.

I can't reproduce the CI failure here but it seems to be consistent there. Are you getting it on your machine?

glts:
Thank you, @technomancy!

The failures are not consistent, though. CI was green at least once, and the error messages vary. I see both `gpg: keyserver receive failed: Cannot assign requested address` and `gpg: keyserver receive failed: No data`. Do you know if CircleCI builds are running in Docker? I wonder if there is some special request throttling or something on some CircleCI nodes? I cannot repro on my machine.

glts:
According to nextcloud/docker#687 it is related to IPv6.

Issue #571, opened 2012-05-11
Set default :checksum policy to :fail.


technomancy:
Not sure why it defaults to :warn. Opinions?

PR #2578, opened 2019-05-17
Defer to Pomegranate for repository policy defaults

glts:
In #571, the Leiningen default checksum policy was set to `:fail` in anticipation of the same default in Pomegranate. Pomegranate has implemented this change years ago, we no longer need to override the Pomegranate defaults in Leiningen.

See #571 for the history.


glts:
Unless we accept the ‘explicit is better than implicit’ argument here …

technomancy:
Sounds good; let's merge it.

PR #2579, opened 2019-05-17
Attempt to fix brittle GnuPG tests in CI

glts:
No idea if this works, but could be worth a try?


glts:
Thank you for the invite, @technomancy. I’m going to merge this to see if it helps with the build failures (until we find something better).

technomancy:
Thanks for all your help identifying stale issues.

glts:
And thank _you_. I just have a little bit of time to spend …

CI is red again, looks like this change was not effective, have to look into it more

Issue #1319, opened 2013-09-08
can not set proxy when use "lein deploy clojars"


yxzhang:
I have tried three methods none of which can work.

(1) in ~/.lein/profiles.clj

{:user 
      {
       :jvm-opts ["-Dhttp.proxyHost=127.0.0.1" "-Dhttp.proxyPort=8087" "-Dhttps.proxyHost=127.0.0.1" "-Dhttps.proxyPort=8087"]
}} 

lein with-profile user  deploy clojars

OR

lein  deploy clojars

(2) 
export http_proxy="http://localhost:8087" 
export https_proxy="http://localhost:8087"

OR

export http_proxy="127.0.0.1:8087" 
export https_proxy="127.0.0.1:8087"

(3)
export LEIN_JVM_OPTS="-Dhttp.proxyHost=127.0.0.1  -Dhttp.proxyPort=8087  -Dhttps.proxyHost=127.0.0.1 -Dhttps.proxyPort=8087"

danielcompton:
One option for testing is to use [`mitmproxy`](https://mitmproxy.org/) as an http proxy on a local machine.

PR #2581, opened 2019-05-21
Also set proxy settings when deploying

glts:
This was prompted by #1319, but I admit I cannot really test this.

The change here simply adds the proxy settings used for dependency resolution also for deployment. According to the Pomegranate Aether docs this is how you do it, I checked that the setting is passed in correctly but I wasn’t able to actually try deploying through a proxy.

### DIFF
--- a/leiningen-core/src/leiningen/core/project.clj
+++ b/leiningen-core/src/leiningen/core/project.clj
@@ -311,7 +311,7 @@
 
 (def deploy-repositories
   (with-meta
-    [["clojars" {:url "https://clojars.org/repo/"
+    [["clojars" {:url "https://repo.clojars.org/"
                  :password :gpg :username :gpg}]]
     {:reduce reduce-repo-step}))
 
--- a/src/leiningen/deploy.clj
+++ b/src/leiningen/deploy.clj
@@ -217,7 +217,8 @@ be able to depend on jars that are deployed without a pom."
                         (:version project)]
           :artifact-map files
           :transfer-listener :stdout
-          :repository [repo])
+          :repository [repo]
+          :proxy (classpath/get-proxy-settings))
          (catch org.eclipse.aether.deployment.DeploymentException e
            (when main/*debug* (.printStackTrace e))
            (main/abort (abort-message (.getMessage e)))))))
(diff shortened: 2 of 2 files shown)


glts:
Good news, I’ve been able to test this with mitmproxy and confirm that this fixes #1319.

glts:
Hold on –

@tobias, in commit 21badb1503274a901ee690101a187b2ea032b857 you updated the Clojars URL from `https://clojars.org/repo/` to `https://repo.clojars.org/`, but only for dependency resolution, not for deployment.

Can you explain why you didn’t update the deploy repository? In this pull request, updating the deploy repository URL helps fix another issue (I’ve checked that it works), so I wonder if there are problems with this change.

tobias:
Hi @glts - when I made that change, the CDN that handles `https://repo.clojars.org/` wasn't configured to support deployment. It has since been configured to handle deployments (essentially proxying them to `https://clojars.org/repo/`), so it should be safe to make this change. However, it's unclear if that change needs to be made to support using the proxy settings. Are you making the change because you noticed the difference, or because it is required for this fix?

glts:
Thank you, @tobias. Both actually, I noticed that the fix here only worked when both `:repositories` and `:deploy-repositories` had the same URL, I suspect it’s an unrelated bug when the repository settings are merged somewhere … but I haven’t investigated.

But if `https://repo.clojars.org/` is fine then it would be easiest make this change here.

glts:
@tobias If you are associated with Clojars this might be interesting for you. Only when I go through mitmproxy, fetching stuff from `https://clojars.org/repo/` leads to a timeout:

    Could not transfer artifact ch.gluet:couplet:jar:0.2.0 from/to clojars (https://clojars.org/repo/): Read timed out

This impacts `lein deploy`, because `lein deploy` first needs to get some Maven metadata from the repository before it can upload anything. So that timeout makes deployment fail. Note that the timeout happens somewhere in Pomegranate/Aether, because I can see that I do eventually get an answer from Clojars, but only after two minutes or so.

This behaviour could even be related to mitmproxy, no idea.

In any case, for proper proxy support we need to align the deploy repository URL, so the pull request as here proposed stands.

glts:
The last question here was if we could align the Clojars repo URLs to `https://repo.clojars.org/`. I don’t see any real obstacles, so if there are no objections I think this is good to go.

tobias:
It's fine to use repo.clojars.org for both uploads and downloads. For the timeout you are seeing, I suspect that's a mitmproxy configuration or other network issue - downloading from clojars.org/repo/ is just ngingx serving files from disk, there's nothing complicated there.

glts:
All right, with this input I think we can go for the merge, thanks.

PR #2582, opened 2019-05-23
Cache ~/.gnupg directory in CI

glts:
GPG tests in CI still occasionally fail, it’s the public keys that sometimes cannot be downloaded.

I don’t know CircleCI, but I imagine if we cache the ~/.gnupg directory, then successful runs will store the downloaded public keys in there.


glts:
I think it’s safe to try this, feel free to revert if not ok

Issue #2580, opened 2019-05-17
‘lein deps :verify’ ignores top-level :checksum and :update policies


glts:
Repository policies specified at the project level are not taken into account when running `lein deps :verify`.

You can work around this by setting policies per-repository, but top-level policies should not be ignored.

```clojure
:checksum :ignore  ; no effect, checksum policy is :fail (default)

:repositories [["clojars" {:url "https://repo.clojars.org/"
                           :checksum :ignore  ; this works
                           }]]
```

This is with Leiningen 2.9.1.

glts:
Ah, this is the same as #1643.

Issue #1643, opened 2014-08-14
lein deps :verify fails if :repositories uses :env/VARIABLE_NAME


thearthur:
If I specify a repo that references enviroment vars lein deps with this exception. It happens regarless of the presence (or absence) of the s3-wagon-private plugin. On all our internal repos it happens with the s3-wagon plugin and happens in the same way with a blank project.

``` sh
arthur@a:~/tmp$ lein new fooo
Generating a project called fooo based on the 'default' template.
The default template is intended for library projects, not applications.
To see other templates (app, plugin, etc), try `lein help new`.
arthur@a:~/tmp$ cd fooo/
arthur@a:~/tmp/fooo$ ls
doc  LICENSE  project.clj  README.md  resources  src  test
arthur@a:~/tmp/fooo$ lein deps :verify
:no-key [cider/cider-nrepl "0.7.0"]
:no-key [cljs-tooling "0.1.3" :exclusions [[org.clojure/clojure]]]
:no-key [compliment "0.1.3" :exclusions [[org.clojure/clojure]]]
:no-key [org.clojure/java.classpath "0.2.0" :exclusions [[org.clojure/clojure]]]
:no-key [org.clojure/tools.namespace "0.2.5" :exclusions [[org.clojure/clojure]]]
:no-key [org.clojure/tools.trace "0.7.8" :exclusions [[org.clojure/clojure]]]
:no-key [org.tcrawley/dynapath "0.2.3" :exclusions [[org.clojure/clojure]]]
:no-key [clojure-complete "0.2.3" :exclusions [[org.clojure/clojure]]]
:no-key [org.clojure/clojure "1.6.0"]
:no-key [org.clojure/tools.nrepl "0.2.3" :exclusions [[org.clojure/clojure]]]
arthur@a:~/tmp/fooo$ emacs project.clj 
arthur@a:~/tmp/fooo$ cat project.clj 
```

``` Clojure
(defproject fooo "0.1.0-SNAPSHOT"
  :description "FIXME: write description"
  :url "http://example.com/FIXME"
  :license {:name "Eclipse Public License"
            :url "http://www.eclipse.org/legal/epl-v10.html"}
  :repositories [["yummly-s3" {:url "s3://XXXXX/releases/"
                               :snapshots {:update :always}
                               :username :env/AWS_KEY 
                               :passphrase :env/AWS_SECRET}]] 
  :dependencies [[org.clojure/clojure "1.6.0"]])
```

``` sh
[...]

PR #2583, opened 2019-05-25
Support repo auth and proxies in ‘lein deps :verify’ too

glts:
In `lein deps :verify`, Pomegranate’s `aether/resolve-dependencies` is called directly without applying repository auth or proxy settings to repositories and mirrors. This change applies those settings there too, and so should provide a fix for both #2580 and #1643.

My implementation feels awkward, so I’m grateful for any feedback. Basically, I extracted the keyword arguments that need to be passed to `aether/resolve-dependencies` into a new public function `leiningen.core.classpath/default-aether-args`, and then used this from `leiningen.deps`. It is as minimal as I could think of.

### DIFF
--- a/leiningen-core/src/leiningen/core/classpath.clj
+++ b/leiningen-core/src/leiningen/core/classpath.clj
@@ -250,6 +250,20 @@
                            checksum (assoc :checksum checksum))]
     [repo-name (merge project-policies opts)]))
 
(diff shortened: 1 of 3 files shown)


technomancy:
This is great. I don't think it's awkward, except in so far as the original code is awkward. (Three separate `let` forms? Really, past-me?) Good work.

PR #2584, opened 2019-05-26
Fix empty/default test selector logic

glts:
This check in `leiningen.test/read-args` was intended to catch unknown or misspelled test selectors, but was never actually triggered:

```clojure
(when (and (empty? selectors)
           (seq given-selectors))
  (main/abort "Please specify :test-selectors in project.clj"))
```

When an unknown selector is passed to `lein test :unknown`, Leiningen silently falls back to the `:default` selector. The change here fixes the logic, `lein test :unknown` now aborts.

### DIFF
--- a/src/leiningen/test.clj
+++ b/src/leiningen/test.clj
@@ -183,14 +183,14 @@
                                             {:only only-form}
                                             (:test-selectors project))
                                      given-selectors)
-        selectors (if (and (empty? selectors)
-                           (:default (:test-selectors project)))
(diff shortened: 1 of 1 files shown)


glts:
This is such a small change, I think there is no harm if I merge it …

Also, looks like CI still has trouble with the GPG tests, I will make one final attempt at a fix.

technomancy:
Nice catch; thanks.

glts:
@technomancy Thanks, and thanks for looking through the other issues.

Issue #1267, opened 2013-07-29
Default :base dependencies get added to pom.xml


alexhall:
The `tools.nrepl` and `clojure-complete` dependencies defined in the :base profile found in `leiningen.core.project/default-profiles` are making their way into the pom.xml file generated by `lein pom`. This can be demonstrated as follows (using a clean install of Leiningen 2.2 and no global profiles defined in ~/.lein):

```
> lein new foo
> cd foo
> lein pom
> grep clojure-complete pom.xml
```

It seems there are a couple of issues at play here:
1. It's unclear to me what the purpose of the `:base` profile is - it's not discussed in the profiles section of the documentation other than to mention that it's included by default. But, as these dependencies are geared towards REPL development, perhaps the `:dev` profile would be a better place for them.
2. The `pom` task unmerges other 'special' profiles (user, provided, dev) but not :base. Is that an oversight - should that also be excluded?

PR #2587, opened 2019-05-30
Remove broken excluded profiles normalization

glts:
I have been puzzling over this for a bit now, and I think this code from 328b2182b08e1f6e3d72d6bf7ccf2847de1b1224 is simply broken.

* In the case where the `exclude-profiles` argument contains keywords, no ‘normalization’ happens at all; the two removed lines are a noop.
* In the (unusual) case where the `exclude-profiles` argument contains profile maps, these get flattened by `mapcat` into map entries. This is not valid data for excluded-profiles.

I don’t understand what the code was meant to do (#1267 is working fine); but since it is ineffective/broken I’m deleting it here.

### DIFF
--- a/leiningen-core/src/leiningen/core/project.clj
+++ b/leiningen-core/src/leiningen/core/project.clj
@@ -887,8 +887,6 @@
                                    (utils/last-distinct-by first))
         include-profiles (map first include-profiles-meta)
         exclude-profiles (utils/last-distinct (expand-profiles project exclude-profiles))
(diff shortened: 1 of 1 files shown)


glts:
The faulty code path is taken only in the very uncommon case where you call `unmerge-profiles` repeatedly with a composite argument, `(-> p (unmerge-profiles [{:a 1}]) (unmerge-profiles [{:a 2}]))`: this throws an exception.

The two lines are relics from an earlier stratum of the profiles code – let’s merge this.

PR #2588, opened 2019-06-05
Small adjustments in profile map setup

glts:
This fixes one unreported problem in the profile map setup and a problem in `meta-merge`, and tidies up a related bit in passing. See the commit messages for details.


glts:
These changes should be Mostly Harmless, I’m merging it.

Issue #2385, opened 2018-01-11
Uberjar creation failing due to snapshot dep in :dev profile


MatthewDarling:
## Reproduction repo

1. [Minimum example project](https://github.com/MatthewDarling/lein-uberjar-fails-because-of-dev-profile)
2. See [this commit specifically](https://github.com/MatthewDarling/lein-uberjar-fails-because-of-dev-profile/commit/f791f4e33d8dbbad16acdd57921bf9075765f9dc) for a diff against the default `lein new` output

## Profile inclusion in uberjars

Leiningen's [profiles documentation](https://github.com/technomancy/leiningen/blob/98b639cb469479c0b79a2ec7e7f9755e338275e2/doc/PROFILES.md#debugging) has a note about profile handling in the `uberjar` task:

> In order to prevent profile settings from being propagated to other projects that depend upon yours, the :default profiles are removed from your project when generating the pom, jar, and uberjar

However, [that remove step](https://github.com/technomancy/leiningen/blob/98b639cb469479c0b79a2ec7e7f9755e338275e2/src/leiningen/uberjar.clj#L178-L179) happens after [the check which prevents released versions from having snapshots](https://github.com/technomancy/leiningen/blob/98b639cb469479c0b79a2ec7e7f9755e338275e2/src/leiningen/uberjar.clj#L164).

This means that if any of the `:default` profiles has a SNAPSHOT dependency it will prevent the creation of a release uberjar. In the example repo for this, it's caused by a dependency in the `:dev` profile of `project.clj`.

## Workaround

While a bare run of `lein uberjar` will crash, `with-profile` can help avoid it. The following two commands both complete successfully on the example repo:

  1. `lein with-profile -dev uberjar`
  2. `lein with-profile uberjar uberjar`

Despite these workarounds, the documentation makes me think it shouldn't be necessary - that the intention is for the `uberjar` task to be isolated from `:dev`, `:user`, and other default profiles.

## Possibly related issues/commentary

1. #960, especially [this comment](https://github.com/technomancy/leiningen/issues/960#issuecomment-12794343) saying:
    > We strip out the default profiles for uberjaring; this is intentional to avoid "leakage of dev settings".
[...]

PR #2596, opened 2019-06-14
Unmerge non-leaky profiles for snapshot dependency check in ‘uberjar’

glts:
In #1845, an early snapshot dependency check was introduced in `uberjar`. The check does not unmerge non-leaky profiles; this results in failure when default profiles such as `:dev` contain snapshot dependencies. The proposed change does the necessary unmerging thereby fixing #2385.


glts:
This should be a harmless change (uberjar itself is not affected, only the early snapshot deps check), let’s merge it.

technomancy:
Makes sense; thanks.

Issue #2552, opened 2019-03-31
Specify secondary license in default LICENSE file?


danielcompton:
https://github.com/technomancy/leiningen/blob/1dc131128279ef50f21d4d643c4d5c2133c1370d/resources/leiningen/new/default/LICENSE#L266 says

> "This Source Code may also be made available under the following 
>Secondary Licenses when the conditions for such availability set forth 
>in the Eclipse Public License, v. 2.0 are satisfied: {name license(s),
>version(s), and exceptions or additional permissions here}."

I assume that this is meant to reference GPL in here? What is the correct language here to line up with the README and project.clj default license?

technomancy:
Ugh; yeah it looks like I didn't read that pull request thoroughly enough. The intention was that it should reference the GPL here.

technomancy:
Yeah, I fixed the reference in project.clj and readme.md but missed that reference in the licens itself I guess https://github.com/technomancy/leiningen/commit/694691f7825ae91fcd89afb369b0d256e91c61a8

technomancy:
Looks like this is fixed.

PR #2600, opened 2019-06-30
Specify secondary license in license files

gavinkflam:
Secondary license should be specified in LICENSE files.

See #2552 as well.


technomancy:
Good catch; thanks.

PR #2601, opened 2019-07-05
Update TUTORIAL.md with {:insecure? true} in Running Code section

HardcoreParkore:
Because without prior configuration, the previous command fails due to:

Syntax error (SunCertPathBuilderException) compiling at (form-init6802313204423860054.clj:1:15).
unable to find valid certification path to requested target


technomancy:
Sorry but this is the wrong fix for the problem; if your JVM is unable to connect to this site then it's important that you get the TLS issues sorted out it rather than working around the problem.

PR #2602, opened 2019-07-08
Add FAQ entry: illegal reflective access

winks:
This came up on IRC last week. Not 100% sure it's the best solution though.


technomancy:
Sounds good; we should definitely address this. I will add a note saying that it's better to use Java 8 if possible too.

PR #2603, opened 2019-07-18
Remove duplicated test project explanation info from test projects

SherSpock:
Justification:
1) There is already an explanatory README present in the root directory
of test_projects
2) About half of the test projects have this duplicated information.
The other half do not, adding inconsistency to duplication.

An alternative:
1) Add the same duplicated info to all test projects.


SherSpock:
Thoughts?

technomancy:
Sure; why not.

Issue #2402, opened 2018-02-14
Atomic/resumable release tasks


danielcompton:
### Problem statement:

When doing releases using `lein release` if the release fails, your project can be left in a partially released state. To fix it you either need to roll back the commits and tags that were made, or roll forward manually, completing the rest of the release by hand. It would be great if there was a way to prevent this from happening.

### Ideas:

* Some kind of dry-run mode like #2229 to allow people to preview what will happen before they release
* Before actually running the release, check that all of the steps can be completed (this is hard to do well because it may depend on remote state like GitHub and Clojars)
* Provide a way to rollback 
* If a release fails, show the user which steps succeeded, which one failed, and which ones are left to run.
* Collect the steps remaining to run in a failed release and store them in a temporary file like `.lein-release-tasks` so they can be rerun on a second attempt.

technomancy:
> When doing releases using lein release if the release fails, your
> project can be left in a partially released state. To fix it you
> either need to roll back the commits and tags that were made, or roll
> forward manually, completing the rest of the release by hand. It would
> be great if there was a way to prevent this from happening.

Atomic releases would be really nice to have, and we've discussed it
before, but the problem is it depends on having atomic deploys. Our
hands are somewhat tied by the existing semantics of Aether. We could
create a new clojars-specific wagon which supported wrapping up all the
files and signatures into a single atomic unit that gets committed all
at once, but last I looked there was no way to do this in the existing
ecosystem without doing a little embrace-and-extend action.

I'd be open to a proposal for this, but it's a lot of work with a lot of
subtle gotchas.

I'm a little more skeptical of the idea of rollbacks considering you can
set release-tasks to any arbitrary task, so it would require a more
general-purpose notion of rolling back that could apply universally. But
if we had atomic deploys plus a "look before you leap" command for git
tagging, I think that'd get you 99% of the way there.

> * If a release fails, show the user which steps succeeded, which one
> failed, and which ones are left to run.

This is a great idea with no downsides as far as I'm concerned. I'd
accept a patch for this in a heartbeat.

> * Collect the steps remaining to run in a failed release and store
> them in a temporary file like .lein-release-tasks so they can be rerun
> on a second attempt.

Likewise for this, assuming it built on the previous step.

PR #2605, opened 2019-07-22
print progress in lein release

saitouena:
# justification
This make it easier to rollback release-tasks or complete them by hand.
https://github.com/technomancy/leiningen/issues/2402

>  * If a release fails, show the user which steps succeeded, which one
>  failed, and which ones are left to run.

`lein release` works like this with this patch.

```
$ cat project.clj 
;; This is Leiningen's own project configuration. See doc/TUTORIAL.md
;; file as well as sample.project.clj for help writing your own.

(defproject leiningen "2.9.1"
...
  :release-tasks [["echo" "aaa"]
                  ["echo" "bbb"]
                  ["echo" "ccc"]]
  :source-paths ["leiningen-core/src" "src"]
  :eval-in :leiningen)
$ bin/lein release
[ 0 / 3 ] Executing lein echo aaa
aaa
[ 1 / 3 ] Executing lein echo bbb
bbb
[ 2 / 3 ] Executing lein echo ccc
ccc
[ 3 / 3 ] Completed lein release.
```

### DIFF
--- a/src/leiningen/release.clj
+++ b/src/leiningen/release.clj
@@ -129,9 +129,12 @@ bump. If none is given, it defaults to :patch."
   ([project] (release project *level*))
   ([project level]
      (binding [*level* (if level (read-string level))]
-       (doseq [task (:release-tasks project)]
-         (let [current-project (project/init-project (project/read))]
-           (main/resolve-and-apply current-project task))))))
+       (let [release-tasks (:release-tasks project)
+             task-count (count release-tasks)]
+         (doseq [[i task] (map vector (range 1 (inc task-count)) release-tasks)]
+           (apply main/info "[" i "/" task-count "] Running lein" task)
+           (let [current-project (project/init-project (project/read))]
+             (main/resolve-and-apply current-project task)))))))
 
 ;; support existing release plugin:
 ;; https://github.com/technomancy/leiningen/issues/1544


src/leiningen/release.clj
```diff
@@ -129,9 +129,13 @@ bump. If none is given, it defaults to :patch."
   ([project] (release project *level*))
   ([project level]
      (binding [*level* (if level (read-string level))]
-       (doseq [task (:release-tasks project)]
-         (let [current-project (project/init-project (project/read))]
-           (main/resolve-and-apply current-project task))))))
+       (let [all-tasks (:release-tasks project)
+             total-task-count (count all-tasks)]
+         (doseq [[i task] (zipmap (range total-task-count) all-tasks)]
```
danielcompton: I don't think `zipmap` is the right function for this, it returns a map, not a sequence. Below 8 entries everything will appear to work, but above 8 the order of the entries in the map is not the same as their insertion. I think `map` would work better here.

Side note, `zipmap` (and `map`) can take a finite and infinite sequence and produce a finite map, e.g. `(zipmap (range) all-tasks)` would also work.

src/leiningen/release.clj
```diff
@@ -129,9 +129,13 @@ bump. If none is given, it defaults to :patch."
   ([project] (release project *level*))
   ([project level]
      (binding [*level* (if level (read-string level))]
-       (doseq [task (:release-tasks project)]
-         (let [current-project (project/init-project (project/read))]
-           (main/resolve-and-apply current-project task))))))
+       (let [all-tasks (:release-tasks project)
+             total-task-count (count all-tasks)]
```
danielcompton: task-count might be a better name?

```suggestion
             task-count (count all-tasks)]
```

src/leiningen/release.clj
```diff
@@ -129,9 +129,13 @@ bump. If none is given, it defaults to :patch."
   ([project] (release project *level*))
   ([project level]
      (binding [*level* (if level (read-string level))]
-       (doseq [task (:release-tasks project)]
-         (let [current-project (project/init-project (project/read))]
-           (main/resolve-and-apply current-project task))))))
+       (let [all-tasks (:release-tasks project)
```
danielcompton:
```suggestion
       (let [release-tasks (:release-tasks project)
```

src/leiningen/release.clj
```diff
@@ -129,9 +129,13 @@ bump. If none is given, it defaults to :patch."
...
+             total-task-count (count all-tasks)]
+         (doseq [[i task] (zipmap (range total-task-count) all-tasks)]
+           (apply main/info (concat ["[" i "/" total-task-count "] Executing lein"] task))
```
danielcompton: I prefer "running" to "executing" but I think this is mostly a matter of taste.

```suggestion
           (apply main/info (concat ["[" i "/" total-task-count "] Running lein"] task))
```

src/leiningen/release.clj
danielcompton: I think we might want to start counting at 1 rather than 0 here? When I see 2/3, I think that that is the second task of the three, when it's actually the last one. That would also mean we could avoid having a final "Completed lein release." log message at the end.

saitouena: > I don't think zipmap is the right function for this, it returns a map, not a sequence.

I should use `(map vector (range) release-tasks)` here.

> I think we might want to start counting at 1 rather than 0 here?
I'm not sure about this. I realized Chrome text search(Ctrl+F) starts from 1, so I'll take your suggestion.

saitouena:
@danielcompton I like all your suggestions. 
lein release looks like this now.
```
$ ./bin/lein release
[ 1 / 3 ] Running lein echo aaa
aaa
[ 2 / 3 ] Running lein echo bbb
bbb
[ 3 / 3 ] Running lein echo ccc
ccc
```

CI failed but `./bin/lein test` succeeded in my local environment. Any ideas?  It seems that CI is unstable for now. https://circleci.com/gh/technomancy/leiningen

src/leiningen/release.clj
```diff
@@ -129,9 +129,12 @@ bump. If none is given, it defaults to :patch."
   ([project] (release project *level*))
   ([project level]
      (binding [*level* (if level (read-string level))]
-       (doseq [task (:release-tasks project)]
-         (let [current-project (project/init-project (project/read))]
-           (main/resolve-and-apply current-project task))))))
+       (let [release-tasks (:release-tasks project)
+             task-count (count release-tasks)]
+         (doseq [[i task] (map vector (range 1 (inc task-count)) release-tasks)]
```
danielcompton: We don’t need the extra `(inc task-count)` range here, map will stop once any of its inputs runs out.

src/leiningen/release.clj
```diff
@@ -129,9 +129,12 @@ bump. If none is given, it defaults to :patch."
...
+           (apply main/info (concat ["[" i "/" task-count "] Running lein"] task))
```
danielcompton: This might be clearer using format, or possibly just removing the concat and wrapping collection, I don’t think it’s necessary?

src/leiningen/release.clj
saitouena: `(range 1)` doesn't work here. Take a look into https://clojuredocs.org/clojure.core/range . `range`'s signature is `(range end)` if it takes only one argument.
I wonder between `(drop 1 (range))` and `(range 1 (inc task-cout))`. The latter is more clear, I think.

src/leiningen/release.clj
saitouena: I agree.

src/leiningen/release.clj
danielcompton: Ah of course, you’re correct, sorry.

technomancy:
Thanks; this looks great. I can't think of any other comparable output to try to make it more consistent with, so I think we should go with this. I'm looking into the CI failure but it's unrelated to your change.

PR #2518, opened 2018-12-31
Add sort for (compilable-namespaces) to provide consistent compile order in all cases. 

LoganGirard:
fixes #2508

Issue #2508, opened 2018-12-16
Wrong compilation order (uberjar, check) (WAS: with linux x86-64)


ikitommi:
Locally everything works fine, but when `lein uberjar` is run with default Clojure Docker container, the compilation order is different and the compilation fails as the Protocols `satisfy?` doesn't match to values. 

I believe the compilation order should be the same (working one) on all platforms?

## lein uberjar

Here's a sample project: https://github.com/ikitommi/uberjar-problem, with guides on how to reproduce this. 

Original problem: https://github.com/metosin/muuntaja/issues/90

## lein check

Saw the same with `lein check`: works locally ok, but not on travis:

* failing run: https://travis-ci.org/metosin/muuntaja/builds/468129603 with this order:

```
Compiling namespace muuntaja.format.msgpack
Compiling namespace muuntaja.interceptor
Compiling namespace muuntaja.middleware
Compiling namespace muuntaja.format.edn
Compiling namespace muuntaja.format.core ;; the defined protocol
Compiling namespace muuntaja.format.json
Compiling namespace muuntaja.format.transit
Compiling namespace muuntaja.parse
Compiling namespace muuntaja.core ;; fails here
```

* successful run: https://travis-ci.org/metosin/muuntaja/builds/447427667 with this order:

```
Compiling namespace muuntaja.core
Compiling namespace muuntaja.format.core ;; the defined protocol
Compiling namespace muuntaja.format.edn
Compiling namespace muuntaja.format.json
Compiling namespace muuntaja.format.transit
Compiling namespace muuntaja.interceptor
Compiling namespace muuntaja.middleware
Compiling namespace muuntaja.parse
Compiling namespace muuntaja.protocols
Compiling namespace muuntaja.util
Compiling namespace muuntaja.format.cheshire
Compiling namespace muuntaja.format.yaml
```

technomancy:
Can you provide a repro case outside docker?

ikitommi:
Only by setting the order manually to the same  as compiled inside docker.

```bash
lein with-profile uberjar/fail uberjar
```

LoganGirard:
So if I'm understanding this issue correctly, the compilation order for namespaces is not identical across platforms?

[...]

PR #2608, opened 2019-07-29
add consistent load order for lein check

saitouena:
related:
- https://github.com/technomancy/leiningen/pull/2518
- https://github.com/technomancy/leiningen/issues/2508

# Justification 
I sometimes came across wierd CI failure at running `lein check`.  This seems like as the same problem of above links.

Please review. I'm not sure about my understanding for this problem. I don't have a minimal repro case for now. I would try to make it if needed.

### DIFF
--- a/src/leiningen/check.clj
+++ b/src/leiningen/check.clj
@@ -9,8 +9,8 @@
   "Check syntax and warn on reflection."
   ([project]
      (let [source-files (map io/file (:source-paths project))
-           nses (b/namespaces-on-classpath :classpath source-files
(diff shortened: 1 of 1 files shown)


saitouena:
It is better to wrap `(sort (b/namespaces-on-classpath ...))` with another function and make sure that call it directly rather than to do like this. Any ideas?

technomancy:
I don't know if this will solve the linked issue, but since we are sorting during compilation, it's sensible to do the same during `check` for consistency. Thanks!

Issue #2572, opened 2019-05-12
ClassCastException when running ‘lein trampoline repl’


glts:
Trying to reproduce #2038 … looks like the nREPL upgrade broke ‘lein trampoline repl’? This used to work in Leiningen 2.8.1.

    $ lein trampoline repl
    ClassCastException java.lang.String cannot be cast to java.lang.Number
            nrepl.server/start-server/addr--1163 (server.clj:120)
            nrepl.server/start-server (server.clj:128)
            nrepl.server/start-server (server.clj:96)
            reply.eval-modes.nrepl/get-connection (nrepl.clj:158)
            reply.eval-modes.nrepl/get-connection (nrepl.clj:156)
            reply.eval-modes.nrepl/main (nrepl.clj:202)
            reply.eval-modes.nrepl/main (nrepl.clj:200)
            reply.main/launch-nrepl/fn--2329 (main.clj:79)
            clojure.core/with-redefs-fn (core.clj:7514)
            clojure.core/with-redefs-fn (core.clj:7498)
            reply.main/launch-nrepl (main.clj:78)
            reply.main/launch-nrepl (main.clj:72)
    Bye for now!

This is with Leiningen 2.9.1.

glts:
This is being talked about in trptcolin/reply#186.

bbatsov:
What exactly is causing Lein to pass the port number as a string? I think it passes the port number as a number pretty much everywhere. Fixing this should be trivial - we just have to identify the offending call in Lein.

glts:
It is a regression within REPL-y, `clojure -Sdeps '{:deps {reply {:mvn/version "0.4.3"}}}' -m reply.main --port 3333` triggers the exception. REPL-y no longer parses the port string.

bbatsov:
I get why this is broken, but I still don't get how come it affects Lein and you patched REPLy before I could check the code there. :-)

PR #2615, opened 2019-08-22
Upgrade reply to 0.4.4

glts:
This upgrades REPL-y to version 0.4.4, fixing #2572.


glts:
This is such a minuscule thing, I say merge.

PR #2616, opened 2019-08-27
Tuned text, added links

dtonhofer:
1) added some links to outside docs; added a link to https://stuartsierra.com/2016/clojure-how-to-ns.html regarding NS discussion
2) tried to improve the text about dash-replaced-by-underscore
3) used output of "tree" instead of "find ." to show filetrees
4) Bumped REPL output examples up to to what is sown in the current version (but didn't change anything regarding the uberjar compilation output, even though it changed a bit)


technomancy:
Thanks for the submission; sorry for the delay.

PR #2617, opened 2019-08-27
Fix #2571: filter the timestamp from pom.properties

### LINKED ISSUE #2571: pom.properties in jar includes timestamp
This prevents jars from being reproducible and should be removed.

It doesn't look like you can fix this with `java.util.Properties`; they always include the date: https://docs.oracle.com/javase/8/docs/api/java/util/Properties.html

Will need to find another way to write these probably?

avfonarev:
The regex might look pretty scary. The timezone string generation is buried in locale code, the rest follows the format of `java.util.Date.toString()`. Works when the comment is `nil` as well (then the first comment line is omitted).

### DIFF
--- a/leiningen-core/src/leiningen/core/project.clj
+++ b/leiningen-core/src/leiningen/core/project.clj
@@ -24,8 +24,11 @@
       (when (.exists git-head)
         (if-let [revision (utils/read-git-head git-head)]
           (.setProperty properties "revision" revision)))
-      (.store properties baos "Leiningen"))
-    (str baos)))
+      (.store properties baos nil))
+    (-> baos
+        str
+        ;; Strip off all the comments printed by java.util.Properties: by default the timestamp is printed
+        utils/strip-properties-comments)))
 
 (defn- warn [& args]
   ;; TODO: remove with 3.0.0
--- a/leiningen-core/src/leiningen/core/utils.clj
+++ b/leiningen-core/src/leiningen/core/utils.clj
@@ -1,6 +1,7 @@
 (ns leiningen.core.utils
   (:require [clojure.java.io :as io]
-            [clojure.java.shell :as sh])
+            [clojure.java.shell :as sh]
+            [clojure.string :as str])
   (:import (com.hypirion.io RevivableInputStream)
            (clojure.lang LineNumberingPushbackReader)
            (java.io ByteArrayOutputStream PrintStream File FileDescriptor
@@ -269,3 +270,9 @@
      (finally
        (System/setErr
         (-> FileDescriptor/err FileOutputStream. PrintStream.)))))
+
+(defn strip-properties-comments
+  "Takes a string containing serialized java.util.Properties
+  and removes all the comment lines (those beginning with #)"
+  [s]
+  (str/replace s #"(\A|\R)(#.+(\R|\z))+" "$1"))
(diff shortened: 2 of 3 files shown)


leiningen-core/src/leiningen/core/project.clj
```diff
@@ -25,7 +25,11 @@
         (if-let [revision (utils/read-git-head git-head)]
           (.setProperty properties "revision" revision)))
       (.store properties baos "Leiningen"))
-    (str baos)))
+    (-> baos
+        str
+        ;; Strip off the timestamp comment printed by java.util.Properties
```
danielcompton: I would also explain why we are stripping the time stamp, namely to make reproducible builds.

leiningen-core/src/leiningen/core/project.clj
```diff
@@ -25,7 +25,11 @@
...
+        (str/replace-first #"(^|\R)#(Sun|Mon|Tue|Wed|Thu|Fri|Sat) (Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec) \d{2} \d{2}:\d{2}:\d{2}.*\d{4}\R" "$1")
```
danielcompton: Might be good to write a few positive and negative tests for this regex so it doesn’t get accidentally broken in the future. Also, nice regex skills, this is very clean.

danielcompton:
> The timezone string generation is buried in locale code

What happens if this is run in a non-English locale?

avfonarev:
Let’s strip off all the comments then. Here is a question though. Since `java.util.Properties` inherits from `Hashtable`, does it guarantee stable iteration order? The entries are spit out using `entrySet().iterator()`. If it doesn’t, this is a more delicate problem than the timestamp comment.

avfonarev:
Well, the question is whether on two different systems one could get both

```
#Leiningen
groupId=ONE
artifactId=props
version=1.0
```
and
```
#Leiningen
version=1.0
artifactId=props
groupId=ONE
```
even if the properties are being set in the same order. Unfortunately, I'm not aware of the hashing implementation details in Java, but I've stumbled upon this [comment on SO](https://stackoverflow.com/a/6184425).

technomancy:
I see, but how is that related to the issue of timestamps?

avfonarev:
Maybe I misunderstand what reproducible builds mean. I thought, the whole point was to make sure that the contents of `pom.properties` must be independent of the system/time of build.

technomancy:
Oh I see what you mean; you're talking about the broader question rather
than the problem of timestamps specifically.

My understanding is that the most basic goals of reproducible builds are
that rebuilding with the same environment (same JDK version, etc) you'll
get the same output. With this consideration, I don't think the ordering
matters because putting the same things into a hashmap with the same
hashing implementation in the same order will give you the same results.

But more broadly, it would be even better if rebuilding with another JDK
gave the same results, and for that purpose it seems like sorting the
lines after they've been rendered to strings is worth doing.

avfonarev:
I've finally found some time to update the code so that all the comments get stripped. Added a test as well. CircleCI is failing for some reason though (can't figure it out from the log).

technomancy:
Not sure what's going on in CI but this works fine; thanks.

PR #2619, opened 2019-09-15
Exclude project-specific `.prepl-port` from version control

rafaeldelboni:
The socket prepl is a relatively new tool built into Clojure that allows you to REPL into a running application. Some new tools like https://github.com/Olical/conjure, use this feature and as nrepl it can generate a file with the current prepl opened port.

This PR just adds ignores for this new file in the templates and in the lein solution.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 .lein*
 /.classpath
 /.nrepl-port
+/.prepl-port
 /.project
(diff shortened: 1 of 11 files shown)


Olical:
I'm all for it 😅

technomancy:
Sure; why not

PR #2621, opened 2019-09-29
Run cljfmt on all files

iyedb:
This is a life saver for those that use editor that run format on save.


technomancy:
Sorry but I don't want to mess up the blame data here, plus these changes aren't always agreeable.

Issue #1621, opened 2014-08-04
Allow disabling of implicit hooks/middleware


technomancy:
It's easier to secure things if there are fewer ways for code to get run.

It's debatable whether auto-loading middleware and hooks were ever a good idea to begin with.

PR #2626, opened 2019-10-23
Apply explicit middleware and hooks even with disabled implicits

### LINKED ISSUE #2625: Declaring :implicit-middleware false appears to turn off explicit middleware too
When listing a plugin's middleware explicitly in the project map and disabling implicit middleware, the explicitly-listed middleware function appears not to be invoked. e.g.

```
:implicit-middleware false
:middleware [foo.plugin/explicit-middleware] ;; this does not get invoked
```

However, when enabling implicit middleware (as default), the same explicitly-listed middleware function does appear to be invoked. e.g.

```
:implicit-middleware true
:middleware [foo.plugin/explicit-middleware] ;; this gets invoked
```

Given that the middleware is listed explicitly (and not named `middleware` in the plugin), it seems it should be invoked regardless of the state of `:implicit-middleware`.

**To Reproduce**

Setup:

* Checkout my MWE repo at [lein-middleware-mwe](https://github.com/radhikalism/lein-middleware-mwe)
  * This contains a trivial plugin (with a non-implicit middleware fn) and two trivial dependent projects only varying in name and `:implicit-middleware` settings

Test with implicit middleware enabled:

* Enter the `lein-middleware-mwe-testproject-with-implicits/` directory
* Execute: `$ lein`
* **Expect:** a single info message line should be printed to console, then the process should immediately exit
* **Actual:** a single info message line is printed to console, then the process immediately exits ✅

Then test with implicit middleware disabled:

* Enter the `lein-middleware-mwe-testproject-without-implicits/` directory
* Execute: `$ lein`
* **Expect:** a single info message line should be printed to console, then the process should immediately exit
* **Actual:** no middleware's info message is printed to console, and the default lein help message is printed instead ❌

**Environment**
- Leiningen Version: 2.9.1, 2.9.0
- Leiningen installation method: manual
- JDK Version:
  - Java 11.0.4 OpenJDK 64-Bit Server VM
  - Java 1.8.0_222 OpenJDK 64-Bit Server VM
- OS: Ubuntu 19.04, Linux 5.0.0-31-generic, x86_64

**Notes**

- The same variant behavior is observable by setting `:implicits` instead of `:implicit-middleware`.
- ~~I tried reproducing this variant behaviour in the leiningen core tests (which doesn't seem to have a relevant test case). But my new test case seems to behave as expected in a test context. It's not clear to me why, or whether main execution follows a different code path.~~
- I may be misunderstanding something about explicit/implicit middleware (this is my first usage), so any clarification is appreciated.
- I have considered not using middleware for my use case.

radhikalism:
To resolve #2625, this change ensures that any explicitly-declared `:middleware` or `:hooks` entries in a project map will still get applied or loaded, even if `:implicits` or `:implicit-middleware` or `:implicit-hooks` is `false`. Only implicitly-loaded middleware or hooks will be skipped.

The bug (if this is a bug) appears to affect all versions since [flags for disabling implicits were introduced](8b9d66cd7c72e3edf5b5a3b69c4e1eb10750a2cf) for #1621, according to a git bisection using the repro steps in #2625.

This is a first cut of a fix, and extends existing tests. Comments welcome.

### DIFF
--- a/leiningen-core/dev-resources/p4.clj
+++ b/leiningen-core/dev-resources/p4.clj
@@ -1,0 +1,4 @@
+(defproject middler-no-implicits "0.0.1"
+  :description "Test some middleware."
+  :middleware [leiningen.core.test.project/add-seven]
+  :implicits false)
--- a/leiningen-core/dev-resources/p5.clj
+++ b/leiningen-core/dev-resources/p5.clj
@@ -1,0 +1,4 @@
+(defproject middler-no-implicit-middleware "0.0.1"
+  :description "Test some middleware."
+  :middleware [leiningen.core.test.project/add-seven]
+  :implicit-middleware false)
--- a/leiningen-core/src/leiningen/core/project.clj
+++ b/leiningen-core/src/leiningen/core/project.clj
@@ -799,27 +799,28 @@
       (utils/error "cannot resolve" hook-name "hook"))))
 
 (defn load-hooks [project & [ignore-missing?]]
-  (when (and (:implicits project true) (:implicit-hooks project true))
-    (doseq [hook-name (concat (plugin-hooks project) (:hooks project))]
-      ;; if hook-name is just a namespace assume hook fn is called activate
-      (let [hook-name (if (namespace hook-name)
-                        hook-name
-                        (symbol (name hook-name) "activate"))]
-        (load-hook hook-name))))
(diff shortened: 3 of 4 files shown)


technomancy:
Thanks for catching this. I think this is a good fix for the problem; thanks for including tests.

PR #2630, opened 2019-11-15
Make sure release tags are annotated

chrisglass:
The git world expects release tags to be of the "annotated" variety.

Currently, tags created with --no-sign are not annotated.

A lot of the surrounding tools rely on this assumption (such as for example "git describe"). This means git describe does not work properly with tags created by leinigen when they are created with --no-sign.

From the git help:

"
Tag objects (created with -a, -s, or -u) are called "annotated" tags; they contain a creation date, the tagger name and e-mail, a tagging message, and an optional GnuPG signature. Whereas a "lightweight" tag is simply a name for an object (usually a commit object).

Annotated tags are meant for release while lightweight tags are meant for private or temporary object labels. For this reason, some git commands for naming objects (like git describe) will ignore lightweight tags by default.
"

### DIFF
--- a/src/leiningen/vcs.clj
+++ b/src/leiningen/vcs.clj
@@ -20,11 +20,13 @@
   (or (:vcs project) (some (partial uses-vcs project) @supported-systems)))
 
 (defn parse-tag-args [args]
-  (loop [parsed-args {:sign? true}
+  (loop [parsed-args {:sign? true :annotate? true}
          args args]
     (case (first args)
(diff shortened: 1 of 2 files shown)


danielcompton:
I think this is a good change. When I worked on `—no-sign` I didn’t realise that turning off signing would make those kinds of tags; it wasn’t intentional on my part. This probably should be called out in release notes as potentially breaking, if people had come to rely on this behaviour.

technomancy:
Good catch; thank you.

PR #2641, opened 2019-12-26
Remove speedbump on upgrade

DavidLiedle:
Problem: I keep my system fresh with a shell script that upgrades various things, including a `lein upgrade`. Every other tool can automate this as a script (rustup, gcloud components, flutter), but `lein upgrade` hangs on a prompt for a "Y" to continue.

This PR proposes cutting out that speedbump and just letting it run through.

### DIFF
--- a/bin/lein
+++ b/bin/lein
@@ -312,33 +312,25 @@ elif [ "$1" = "upgrade" ] || [ "$1" = "downgrade" ]; then
     else
         TARGET_VERSION="${2:-stable}"
         echo "The script at $SCRIPT will be upgraded to the latest $TARGET_VERSION version."
-        echo -n "Do you want to continue [Y/n]? "
-        read RESP
-        case "$RESP" in
-            y|Y|"")
-                echo
-                msg "Upgrading..."
-                TARGET="/tmp/lein-${$}-upgrade"
-                if $cygwin; then
-                    TARGET=$(cygpath -w "$TARGET")
-                fi
-                LEIN_SCRIPT_URL="https://github.com/technomancy/leiningen/raw/$TARGET_VERSION/bin/lein"
-                $HTTP_CLIENT "$TARGET" "$LEIN_SCRIPT_URL"
-                if [ $? == 0 ]; then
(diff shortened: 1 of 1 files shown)


technomancy:
Why not just `echo Y | lein upgrade` instead?

I don't think it's a good idea to remove this altogether. Maybe a flag like `--yes` or something instead if the pipe doesn't work.

DavidLiedle:
I wanted to start exactly this dialogue. I would be content to strip it, but we could add a flag parsing segment to the whole shebang here. Happy to work in the direction desired by the community. My low-hanging-fruit solution is viable: who needs that typo-check if they're updating? Should we all pause, confirm that we do, indeed, wish to update, and continue, or should we add flags?

DavidLiedle:
Follow-up observation: Many CLI's do, in fact, go with a `-y` (apt comes to mind) to plow through. Happy to support whatever is going on outside of my bubble. Not so much work to BASH it in, but a definite decision to be made here.

DavidLiedle:
BTW - my CI failed due to an IRC com failure, which absolutely delights me... Where can I see the config to ping IRC on build from GitHub? <3

technomancy:
> who needs that typo-check if they're updating?

The rationale is that upgrading can be a pretty drastic change--we don't know what's in the new version, so we want to make sure the user understands what's going on here. It's literally replacing the program that's currently in use, which is much higher-impact than any other `lein` command that exists.

> CI failed due to an IRC com failure, which absolutely delights me... Where can I see the config to ping IRC on build from GitHub? <3

Sadly the IRC feature has been broken for at least a couple years. =( Just never got around to removing the config for it. That's not the cause of the CI failure though.

So yeah, I think the best thing to do is to skip the check only when a `-y` or `--yes` flag is detected; thanks.

glts:
Just to make sure you’re aware of this option, haven’t seen it mentioned. Unixes have the yes command for precisely this:

    yes | lein upgrade

technomancy:
Going to go ahead and close this out since I think `yes` is the better solution.

PR #2647, opened 2020-01-13
Include PGP public keys to use in tests

glts:
Don’t know if you’re aware of this, but CI doesn’t look healthy and it’s
all due to one GnuPG test I added a while ago. Apparently in the CI
environment there are issues with fetching public keys from key servers.

Finally, here is a fix that simply avoids fetching the key. Unless there
are objections I plan to merge this so that we’re finally back to an
always-green CI. Cheers.


technomancy:
That looks great; thanks for looking into this. I haven't been following the CI status since the IRC notifications have been broken.

Issue #2649, opened 2020-01-28
Publish new version of lein-pprint with no-pretty option


robhanlon22:
Looks like an option was added to `lein-pprint` in 2018 that allows for unformatted printing: https://github.com/technomancy/leiningen/commit/454c3dc640f4b3507b13154bad2ee8ff91d702c9 However, this change was never published. Could the version of `lein-pprint` be bumped and pushed to Clojars? Thanks.

technomancy:
Sure thing; pushed this and bumped to 1.3.0 bc368fa

robhanlon22:
Thx!

PR #2651, opened 2020-01-31
Fix broken pprint plugin

robhanlon22:
I requested that the version of `lein-pprint` including `--no-pretty` be published in #2649. Turns out that it's broken in a couple of ways! This PR fixes it to work, and makes the docs consistent with the code.


technomancy:
Cool; thanks. Will get this deployed soon.

technomancy:
This is live as 1.3.2.

PR #2654, opened 2020-02-09
Bump pprint version to match Clojars, improve docs

robhanlon22:


technomancy:
Wups, yeah forgot to push. Good catch.

PR #2661, opened 2020-03-04
Memoize application of middleware

c2nes:
This change introduces memoization of middleware. This is meant to improve performance when using middleware that is expensive or slow to apply. The performance improvement is particularly significant when running tasks like "install", "jar" or "pom" which involve repeated merging and unmerging of profiles, as both of these operations involve reapplying middleware.

To avoid issues with embedded use in long running applications, this memoization is only enabled by default when running `lein` on the command line. Memoization can also be disabled by specifying `:memoize-middleware? false` in the project map or user profile.

### DIFF
--- a/leiningen-core/src/leiningen/core/main.clj
+++ b/leiningen-core/src/leiningen/core/main.clj
@@ -443,13 +443,14 @@ Get the latest version of Leiningen at https://leiningen.org or by executing
     (project/ensure-dynamic-classloader)
     (aether/register-wagon-factory! "http" insecure-http-abort)
     (user/init)
-    (let [project (if (.exists (io/file *cwd* "project.clj"))
-                    (project/read (str (io/file *cwd* "project.clj")))
-                    (default-project))]
-      (when (:exact-lein-version project) (verify-exact-version project))
-      (when (:min-lein-version project) (verify-min-version project))
-      (configure-http)
-      (resolve-and-apply project raw-args))
+    (binding [project/*memoize-middleware* true]
+      (let [project (if (.exists (io/file *cwd* "project.clj"))
+                      (project/read (str (io/file *cwd* "project.clj")))
+                      (default-project))]
+        (when (:exact-lein-version project) (verify-exact-version project))
+        (when (:min-lein-version project) (verify-min-version project))
+        (configure-http)
+        (resolve-and-apply project raw-args)))
     (catch Exception e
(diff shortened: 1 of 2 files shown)


leiningen-core/src/leiningen/core/main.clj
```diff
@@ -443,13 +443,14 @@ Get the latest version of Leiningen at https://leiningen.org or by executing
...
-      (configure-http)
-      (resolve-and-apply project raw-args))
+    (binding [project/*memoize-middleware* true]
+      (let [project (if (.exists (io/file *cwd* "project.clj"))
+                      (project/read (str (io/file *cwd* "project.clj")))
+                      (default-project))]
+        (when (:exact-lein-version project) (verify-exact-version project))
+        (when (:min-lein-version project) (verify-min-version project))
+        (configure-http)
+        (resolve-and-apply project raw-args)))
```
technomancy: This looks good, but have you considered something more like this instead?

```clj
(binding [project/*memoize-middleware*
          (:memoize-middleware? (:without-profiles (meta project) project)
                                (:memoize-middleware? (:user (user/profiles))
                                                      true))]
  (resolve-and-apply project raw-args))
```

It seems like if we just figure this out once at the entry point then a lot of the stuff further down in this patch can be removed, unless you already considered this and decided against it for reasons I'm not seeing. =)

c2nes: I actually hadn't considered it, but I could see a couple potential consequences,

1. The initial call to `project/init-project` would run without memoization enabled. This would mean that in some cases middleware might execute an extra time. It would also mean that middleware that read other projects (e.g. lein-modules and lein-parent) would run their first time without the benefit of memoization.
2. When other projects are read (as with lein-modules/lein-parent) the `:memoize-middleware?` setting for these projects will be ignored with the setting for the "root" project being respected instead. 

In the case of the second point I'm not if this is a problem or an improvement 😄 

I think the first consequence would be a bit unfortunate, but could also be avoided if we were to call `read-raw` and `init-project` directly, with `*memoize-middleware*` initialized in between.

technomancy: Ah, at first I thought that point 2 would be a good thing, (if the plugin wants memoization, it should bind it in the plugin code) but then I realized that this would be very difficult to do in a backwards-compatible way. Let's go with what you've got here then!

PR #2670, opened 2020-03-22
remove deprecated make-pom-properties

saitouena:
`make-pom-properties` have never been working since 2014.
`make-pom-properties` was calling `project/make-project-properties`
without arguments while `project/make-project-properties` requires one argument.
So it's time to remove this.


technomancy:
Haha; nice catch. Thanks.

PR #2676, opened 2020-04-17
fix race condition in utils/mkdirs

VaclavSynacek:
There is a race condition in `mkdirs` that it tests for existence of directory first and then tries to create it. But if the directory is created between the test and the attempt at creation (by some other process) exception would be thrown even though there is no reason for it as the directory now does exist.

Normally this is rare to encounter. But if you run several `lein run ...` in parallel on one machine, it can occasionally crash.

In my case, a parallelized CI pipeline failed with exception once in about 20 runs.


technomancy:
Good catch; thanks.

PR #2677, opened 2020-04-21
Makes the default JVM options version aware

samuelhnrq:
- Moves up the java cmd declaration
- Conditionally adds the java 13 depreceated `-Xverify:none` on environments with lower java versions.
- Fix the JVM memory flags of lein-pkg
- Fix missing **"** on lein-pkg

### DIFF
--- a/bin/lein
+++ b/bin/lein
@@ -165,9 +165,24 @@ while [ -h "$SCRIPT" ] ; do
     fi
 done
 
+if [ ! -x "$JAVA_CMD" ] && ! type -f java >/dev/null
+then
+    msg "Leiningen couldn't find 'java' executable, which is required."
+    msg "Please either set JAVA_CMD or put java (>=1.6) in your \$PATH ($PATH)."
+    exit 1
+fi
+
+export LEIN_JAVA_CMD="${LEIN_JAVA_CMD:-${JAVA_CMD:-java}}"
+
+DEF_LEIN_JVM_OPTS="-XX:+TieredCompilation -XX:TieredStopAtLevel=1"
+JAVA_VERSION=$($LEIN_JAVA_CMD -version 2>&1 | awk -F '"' '/version/ {split($2, v, "."); print v[1] }')
+if [[ $JAVA_VERSION -lt 13 ]]; then
+    DEF_LEIN_JVM_OPTS+=" -Xverify:none"
+fi
+
 BIN_DIR="$(dirname "$SCRIPT")"
 
-export LEIN_JVM_OPTS="${LEIN_JVM_OPTS-"-Xverify:none -XX:+TieredCompilation -XX:TieredStopAtLevel=1"}"
+export LEIN_JVM_OPTS="${LEIN_JVM_OPTS-$DEF_LEIN_JVM_OPTS}"
 
 # This needs to be defined before we call HTTP_CLIENT below
 if [ "$HTTP_CLIENT" = "" ]; then
@@ -251,15 +266,6 @@ else # Not running from a checkout
         self_install
     fi
 fi
-
-if [ ! -x "$JAVA_CMD" ] && ! type -f java >/dev/null
-then
-    msg "Leiningen couldn't find 'java' executable, which is required."
-    msg "Please either set JAVA_CMD or put java (>=1.6) in your \$PATH ($PATH)."
-    exit 1
-fi
-
-export LEIN_JAVA_CMD="${LEIN_JAVA_CMD:-${JAVA_CMD:-java}}"
 
 if [[ -z "${DRIP_INIT+x}" && "$(basename "$LEIN_JAVA_CMD")" == *drip* ]]; then
     export DRIP_INIT="$(printf -- '-e\n(require (quote leiningen.repl))')"
--- a/bin/lein-pkg
+++ b/bin/lein-pkg
@@ -35,10 +35,21 @@ for f in "/etc/leinrc" "$LEIN_HOME/leinrc" ".leinrc"; do
     fi
 done
 
-export LEIN_JVM_OPTS="${LEIN_JVM_OPTS-"-Xverify:none -XX:+TieredCompilation -XX:TieredStopAtLevel=1"}"
+# Which Java?
 
-grep -E -q '^\s*:eval-in\s+:classloader\s*$' project.clj 2> /dev/null &&
-LEIN_JVM_OPTS="${LEIN_JVM_OPTS:-'-Xms64m -Xmx512m'}"
+export JAVA_CMD="${JAVA_CMD:-"java"}"
+export LEIN_JAVA_CMD="${LEIN_JAVA_CMD:-$JAVA_CMD}"
+
+DEF_LEIN_JVM_OPTS="-XX:+TieredCompilation -XX:TieredStopAtLevel=1"
+JAVA_VERSION=$($LEIN_JAVA_CMD -version 2>&1 | awk -F '"' '/version/ {split($2, v, "."); print v[1] }')
+echo Java version $JAVA_VERSION
+if [[ $JAVA_VERSION -lt 13 ]]; then
+    DEF_LEIN_JVM_OPTS+=" -Xverify:none"
+fi
+export LEIN_JVM_OPTS="${LEIN_JVM_OPTS-$DEF_LEIN_JVM_OPTS}"
+
+grep -E -q '^\s*:eval-in\s+:classloader\s*$' project.clj 2> /dev/null && \
+    LEIN_JVM_OPTS="$LEIN_JVM_OPTS -Xms64m -Xmx512m"
 
 # If you're not using an uberjar you'll need to list each dependency
 # and add them individually to the classpath/bootclasspath as well.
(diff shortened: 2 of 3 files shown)


bin/lein
```diff
@@ -165,9 +165,24 @@ while [ -h "$SCRIPT" ] ; do
...
+    msg "Leiningen couldn't find 'java' executable, which is required."
+    msg "Please either set JAVA_CMD or put java (>=1.6) in your \$PATH ($PATH)."
+    exit 1
+fi
+
+export LEIN_JAVA_CMD="${LEIN_JAVA_CMD:-${JAVA_CMD:-java}}"
+
+DEF_LEIN_JVM_OPTS="-XX:+TieredCompilation -XX:TieredStopAtLevel=1"
+JAVA_VERSION=$($LEIN_JAVA_CMD -version 2>&1 | awk -F '"' '/version/ {split($2, v, "."); print v[1] }')
+if [[ $JAVA_VERSION -lt 13 ]]; then
```
samuelhnrq, line 179: During testing I remembered that java 9 and lower used 1.x versioning but I want to detect 13+ which uses semantic versioning, so this works anyway (1 < 13)

bin/lein-pkg
```diff
@@ -35,10 +35,21 @@ for f in "/etc/leinrc" "$LEIN_HOME/leinrc" ".leinrc"; do
...
+echo Java version $JAVA_VERSION
+if [[ $JAVA_VERSION -lt 13 ]]; then
+    DEF_LEIN_JVM_OPTS+=" -Xverify:none"
+fi
+export LEIN_JVM_OPTS="${LEIN_JVM_OPTS-$DEF_LEIN_JVM_OPTS}"
 
-grep -E -q '^\s*:eval-in\s+:classloader\s*$' project.clj 2> /dev/null &&
-LEIN_JVM_OPTS="${LEIN_JVM_OPTS:-'-Xms64m -Xmx512m'}"
+grep -E -q '^\s*:eval-in\s+:classloader\s*$' project.clj 2> /dev/null && \
+    LEIN_JVM_OPTS="$LEIN_JVM_OPTS -Xms64m -Xmx512m"
```
samuelhnrq, line 52: Took the liberty, not sure If the old behavior was on purpose...

samuelhnrq:
Didn't add the same behavior to the windows scripts (.bat/.ps1) because they did not even had the default flags, I could add them, but I didn't want to deviate the scope of this PR any further.

technomancy:
Thanks for the patch. However, on my system this adds 0.3 seconds to the startup time. I'm not sure this is a good tradeoff for a warning that doesn't affect actual functionality given that startup time is the number one complaint people have?

winks:
The number one complaint people have about leiningen (and Clojure in general, but this has improved) is the slow startup time. AFAIK that's the main reason we haven't implemented this straight-forward idea in the past.

nihilismus:
One option to get java' s version without its execution can be from the (probably existing) file `release` inside `$JAVA_HOME`:

In my system, with something like `find $HOME/.jenv/versions -maxdepth 2 -iname release | sort | while read line; do echo $line; grep _VERSION $line; done`:

```console
$HOME/.jenv/versions/adoptopenjdk11/release
IMPLEMENTOR_VERSION="AdoptOpenJDK"
JAVA_VERSION="11.0.7"
JAVA_VERSION_DATE="2020-04-14"

$HOME/.jenv/versions/adoptopenjdk13/release
IMPLEMENTOR_VERSION="AdoptOpenJDK"
JAVA_VERSION="13.0.2"
JAVA_VERSION_DATE="2020-01-14"

$HOME/.jenv/versions/adoptopenjdk8/release
JAVA_VERSION="1.8.0_252"
OS_VERSION="2.6"

$HOME/.jenv/versions/graalvm19/release
GRAALVM_VERSION=19.3.1

$HOME/.jenv/versions/jbr11/release
JAVA_VERSION="11.0.6"

$HOME/.jenv/versions/zulu11/release
IMPLEMENTOR_VERSION="Zulu11.39+15-CA"
JAVA_VERSION="11.0.7"
JAVA_VERSION_DATE="2020-04-14"

$HOME/.jenv/versions/zulu13/release
IMPLEMENTOR_VERSION="Zulu13.31+11-CA"
JAVA_VERSION="13.0.3"
JAVA_VERSION_DATE="2020-04-14"

$HOME/.jenv/versions/zulu8/release
JAVA_VERSION="1.8.0_252"
OS_VERSION="2.6"

```

Another option can be `strings` [1], but not available in Windows(r).

In my system, with something like `strings $LEIN_JAVA_CMD | grep -E '^[1-9].*\..*'`:

```console
$HOME/.jenv/versions/adoptopenjdk11/bin/java
11.0.7+10

$HOME/.jenv/versions/adoptopenjdk13/bin/java
13.0.2+8

$HOME/.jenv/versions/adoptopenjdk8/bin/java
1.8.0_252-b09

$HOME/.jenv/versions/graalvm19/bin/java
11.0.6+9-jvmci-19.3-b07

$HOME/.jenv/versions/jbr11/bin/java
11.0.6+8-b765.40

$HOME/.jenv/versions/zulu11/bin/java
11.0.7+10-LTS

$HOME/.jenv/versions/zulu13/bin/java
13.0.3+3-MTS

$HOME/.jenv/versions/zulu8/bin/java
1.8.0_252-b14
```

[1] https://sourceware.org/binutils/docs/binutils/strings.html

samuelhnrq:
> Thanks for the patch. However, on my system this adds 0.3 seconds to the startup time. I'm not sure this is a good trade off for a warning that doesn't affect actual functionality given that startup time is the number one complaint people have?

Well I completely agree, I'm just surprised how such a simple addition could possible add that much to the startup time... Maybe my implementation could be improved...

It only adds 30\~57ms here with the awk, the test, and java invocation... Don't know if my CPU is that incredible (Ryzen 5 3600) or my rolling release distro (~_I use arch btw_~)
```bash
$ time [[ $(java -version 2>&1 | awk -F '"' '/version/ {split($2, v, "."); print v[1] }') -lt 13 ]]

real	0m0,044s
user	0m0,036s
sys	0m0,016s
```

But, I bet the cause might be something else. Whilst writing this I found out the flag haven't been truly disabled, [here](https://bugs.openjdk.java.net/browse/JDK-8214731) is the bug which originated the depreciation. What it means is that `-Xverify:none` still is working just now flagged as _'plz dont use'_. And I understand if you guys want to keep it even tough it might be actually removed in the future, precious 300ms amiright? Feel free to close as wontfix. Thanks for your time.

samuelhnrq:
Ok I think my last comment might have a little too much sass, sorry.I understand this 'fix' might seem a bad trade off, 300ms more startup time t

winks:
@nihilismus We don't set or use `JAVA_HOME` in `lein` and I'm not even sure how widespread it's use is.

This my laptop on Ubuntu 18.04, no fiddling around, so I'd say this is a typical installation:

```
$ dpkg -l | grep jdk
rc  openjdk-11-jre-headless:amd64                 11.0.6+10-1ubuntu1~18.04.1                          amd64        OpenJDK Java runtime, using Hotspot JIT (headless)
ii  openjdk-8-jdk-headless:amd64                  8u242-b08-0ubuntu3~18.04                            amd64        OpenJDK Development Kit (JDK) (headless)
ii  openjdk-8-jre-headless:amd64                  8u242-b08-0ubuntu3~18.04                            amd64        OpenJDK Java runtime, using Hotspot JIT (headless)
$ java -version
openjdk version "1.8.0_242"
OpenJDK Runtime Environment (build 1.8.0_242-8u242-b08-0ubuntu3~18.04-b08)
OpenJDK 64-Bit Server VM (build 25.242-b08, mixed mode)
$ env | grep JAVA
```

JAVA_HOME is not set.

winks:
@samuelhnrq 300ms might be much or not much, depending on your view point. I don't disagree that it can be "not much". But as I wrote above, startup time is literally the number 1 complaint about leiningen we've gotten from users, consistently over the years. Many hours by many people were spent analyzing and trying to bring it down, that's why it is indeed a big deal.

samuelhnrq:
> @nihilismus We don't set or use `JAVA_HOME` in `lein` and I'm not even sure how widespread it's use is.

``` which java ```

> @samuelhnrq 300ms might be much or not much, depending on your view point. I don't disagree that it can be "not much". But as I wrote above, startup time is literally the number 1 complaint about leiningen we've gotten from users, consistently over the years. Many hours by many people were spent analyzing and trying to bring it down, that's why it is indeed a big deal.

Oh no I do think 300 ms are a big deal. it's just that is 300 ms that were deprecated and (again IMHO) needs to be removed.

winks:
> > @nihilismus We don't set or use `JAVA_HOME` in `lein` and I'm not even sure how widespread it's use is.
> 
> `readlink -f $(which java)`

That doesn't help. I meant the literal variable, not the concept of where the java binary is located.

Look at my example. There simply *is* no `release` file, so in this case I don't care where the java binary lives.

dan-mcdonald:
Is `-Xverify:none` ever necessary even on earlier JVM? The [PR that added it](https://github.com/technomancy/leiningen/pull/2319) seemed to suggest it was only needed because clojure 1.9 had not been released yet.

technomancy:
I think this is no longer needed with b025ecf.

PR #2678, opened 2020-04-26
Apply test selectors to test vars only

mourjo:
If test selectors select no test var but there is another var in the namespace, then the test namespace is considered and sent to `clojure.test/run-tests` causing once fixtures to run. Since `clojure.test/run-tests` only consider vars with the `:test` metadata, Leiningen could also use that only to select tests.


technomancy:
Good idea; thanks for the patch!

mourjo:
Thanks!

PR #2686, opened 2020-05-29
Add direct-linking jvm-opt to app uberjar profile

### LINKED ISSUE #2684: Should new app template uberjar profile use direct linking by default?
Hello,
I was wondering what's the reason reason new app template uberjar profile doesn't use direct linking by default. It should provide better performance and better startup speed. Seems like an overall benefit.
What do you think?

Thanks
Ben

bsless:
Closes #2684

### DIFF
--- a/resources/leiningen/new/app/project.clj
+++ b/resources/leiningen/new/app/project.clj
@@ -6,4 +6,5 @@
   :dependencies [[org.clojure/clojure "1.10.1"]]
   :main ^:skip-aot {{namespace}}
   :target-path "target/%s"
-  :profiles {:uberjar {:aot :all}})
+  :profiles {:uberjar {:aot :all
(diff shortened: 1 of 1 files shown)


bsless:
There's one failure in the CI and the message says someone should be shouted at?
Is it something I did?
```
FAIL in (clojars-ssl-cert) (keys.clj:13)
If this fails, yell at tcrawley to update the clojars.pem for use by lein as a client cert.
expected: (< (System/currentTimeMillis) (.getTime month-before-cert-expiry))
  actual: (not (< 1590769216445 1589673600000))
```

technomancy:
Haha, no; it's not anything to do with you. That test is there to remind us to keep our TLS certs updated, but it's probably not needed any more. I think it was written before letsencrypt allowed all that stuff to be automated.

@tobias should we get rid of that test?

PR #2688, opened 2020-06-13
add sha256 command for OpenBSD

310b8fb5:


310b8fb5:
The sha256 tool on OpenBSD is `sha256`. This is a simple fix intended to allow lein to run without a `sha256sum` wrapper.

technomancy:
Makes sense; thanks.

Issue #2328, opened 2017-10-19
Unreadable forms in certain eval-in-project calls


technomancy:
Sometimes `eval-in-project` calls will result in forms that contain unreadable `#object...` elements.

I have only been able to reproduce the problem with `lein eastwood` but others report problems with built-in tasks or with `lein ring`.

In the instance with eastwood, somehow the entire project map is being attached as metadata to the form being evaluated, and the metadata contains merge directives like `reduce-dep-step` which is a function.

ccann:
I'm not sure if it's the same issue, but both eastwood and bikeshed fail for me with the same `Caused by: java.lang.RuntimeException: No reader function for tag ...` when compiling the `target` directory of the project.

I also get this line:
`Java HotSpot(TM) 64-Bit Server VM warning: Unable to open cgroup memory limit file /sys/fs/cgroup/memory/memory.limit_in_bytes (No such file or directory)`

pmonks:
<digression warning>
The `Java HotSpot(TM) 64-Bit Server VM warning: Unable to open cgroup memory limit file /sys/fs/cgroup/memory/memory.limit_in_bytes (No such file or directory)` warning is unrelated - it's coming from the JVM, and is a side effect of the fix for https://github.com/technomancy/leiningen/issues/2304.  While quite annoying, my gut tells me that that fix is "right", and that instead the JVM should not be complaining so loudly when it can't read the cgroups info.

technomancy:
@pmonks if that path is predictable we could check for its existence before adding the flag to the `:jvm-opts`.

jafingerhut:
I have done a tiny bit of investigation, and at least so far it appears that the reduce-dep-step things with `#object` tags are there both with Leiningen 2.7.1, and also with Leiningen 2.8.0.  If it would be helpful, I can release to Clojars a version of Eastwood that enables dumping the project map that it passes to `eval-in-project`, showing metadata as well as the normal data.

technomancy:
@jafingerhut thanks; I can just locally install a snapshot version which has what I need. I'm just confused as to why certain uses of `eval-in-project` trigger this and others don't; eastwood doesn't appear to be doing anything unusual with the form being evaluated.

jafingerhut:
[...]

PR #2696, opened 2020-07-23
Fix reflection warning

### LINKED ISSUE #2695: Reflector/invokeStaticMethod reflection warning
**Initial debugging steps**
Before creating a report, _especially_ around exceptions being thrown when running Leiningen, please check if the error still occurs after:

- [x] Updating to using the latest released version of Leiningen (`lein upgrade`).
- [x] Moving your `~/.lein/profiles.clj` (if present) out of the way. This contains third-party dependencies and plugins that can cause problems inside Leiningen.
- [x] Updating any old versions of plugins in your `project.clj`, especially if the problem is with a plugin not working. Old versions of plugins like nREPL and CIDER (as well as others) can cause problems with newer versions of Leiningen.
- [x] (If you are using Java 9 or newer), updating your dependencies to their most recent versions. Recent JDK's have introduced changes which can break some Clojure libraries.

**Describe the bug**
Reflection warning at
https://github.com/technomancy/leiningen/blob/1dc131128279ef50f21d4d643c4d5c2133c1370d/src/leiningen/run.clj#L78-L80

**To Reproduce**
1. lein new app foo
2. add `:global-vars {*warn-on-reflection* true}` to project.clj
3. lein run will print reflection warning

**Actual behavior**
```
Reflection warning, /tmp/form-init8395418479029823401.clj:1:870 - call to static method invokeStaticMethod on clojure.lang.Reflector can't be resolved (argument types: unknown, java.lang.String, unknown).
```

**Expected behavior**
No reflection warning

**Environment**
- Leiningen Version: Leiningen 2.9.4
- Leiningen installation method: manual
- JDK Version: openjdk version "1.8.0_252"
- OS: Ubuntu 20.04
- Anything else that might be relevant to your problem?

**Additional context**
I noticed that people have attempted to fix this before in #1817 and #2306, but it persisted.

valerauko:
Fixes #2695 

* added type hint to the invokeStaticMethod call
* added `*print-meta* true` a'la #2306 

Tests are passing.

### DIFF
--- a/leiningen-core/src/leiningen/core/eval.clj
+++ b/leiningen-core/src/leiningen/core/eval.clj
@@ -239,7 +239,8 @@
                     (io/file (:target-path project) (str checksum "-init.clj")))]
     (spit init-file
           (binding [*print-dup* *eval-print-dup*
-                    *print-meta* true]
+                    ;; only to resolve reflection warning in run-form (run.clj)
+                    *print-meta* (not (empty? (filter #(:lein-with-meta (meta %)) form)))]
             (pr-str (when-not (System/getenv "LEIN_FAST_TRAMPOLINE")
                       `(.deleteOnExit (File. ~(.getCanonicalPath init-file))))
                     form)))
--- a/leiningen-core/src/leiningen/core/eval.clj
+++ b/leiningen-core/src/leiningen/core/eval.clj
@@ -238,7 +238,8 @@
                     (File/createTempFile "form-init" ".clj")
                     (io/file (:target-path project) (str checksum "-init.clj")))]
     (spit init-file
-          (binding [*print-dup* *eval-print-dup*]
+          (binding [*print-dup* *eval-print-dup*
+                    *print-meta* true]
             (pr-str (when-not (System/getenv "LEIN_FAST_TRAMPOLINE")
                       `(.deleteOnExit (File. ~(.getCanonicalPath init-file))))
                     form)))
@@ -321,7 +322,8 @@
                                :port (Integer. (slurp port-file)))
             client (client-session (client transport Long/MAX_VALUE))
             pending (atom #{})]
-        (message client {:op "eval" :code (binding [*print-dup* *eval-print-dup*]
+        (message client {:op "eval" :code (binding [*print-dup* *eval-print-dup*
+                                                    *print-meta* true]
                                             (pr-str form))})
         (doseq [{:keys [out err status session] :as msg} (repeatedly
                                                           #(recv transport 100))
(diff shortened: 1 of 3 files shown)


glts:
No, I don’t see how this is different from #2306. I just tried it and it breaks in the same way.

The last time this was attempted – and released – it caused a lot of damage (sorry). The next attempt needs to have more extensive testing. See #2328 and the mentioned issues there.

valerauko:
It's not different. I thought it was unintentionally removed at some point. I expected there to be tests for any known regressions like that...

If you have a repro I could add to the lein test suite please advise. Currently I can only repro by running `lein eastwood`.

valerauko:
I added a hacky fix using metadata to tag the form in question. Stuff without the magical meta tag won't have metadata printed (this is current behavior). I added a test case for this as well.

The last test failure seems to be a maven connection refused error... I can't rerun the tests on Circle to check if time resolves it

glts:
I don’t remember what the problem was, aren’t all projects with plugins affected? Anyway, I cannot look into this further.

frenchy64:
Perhaps we can always use `*print-meta* true` to generate forms, but first prepare the form by removing all non-:tag metadata (and assert that all tags are symbols or strings) via a postwalk. That might help avoid the problem of generating unreadable forms.

frenchy64:
After some investigation, my idea isn't going to work because you can't remove metadata from vars and other such objects with mutable metadata.

I'll experiment with hybrid of these approaches (use a flag like in this PR to signal a very strict use of `print-meta`).

leiningen-core/src/leiningen/core/eval.clj
```diff
@@ -238,7 +238,10 @@
                     (File/createTempFile "form-init" ".clj")
                     (io/file (:target-path project) (str checksum "-init.clj")))]
     (spit init-file
-          (binding [*print-dup* *eval-print-dup*]
+          (binding [*print-dup* *eval-print-dup*
+                    ;; only to resolve reflection warning in run-form (run.clj)
+                    *print-meta* (and (seq? form)
+                                      (first (filter #(:lein-with-meta (meta %)) form)))]
```
frenchy64, line 244: I'm not sure this is going to have the desired effect because most calls to `shell-command` wrap their arguments with extra forms. For example, see impl of `eval-in-project` and `leiningen.trampoline`'s use of `shell-command`.

frenchy64: I have the feeling this approach is still too dangerous because there might be var literals in a project's `:global-vars` or `:injections`, and they contain unreadable Namespace objects (assuming my understanding is correct that most calls to `shell-command` will come via `eval-in-project`).

PR #2700, opened 2020-08-20
self-install bronked

ViviCetus:
tries to download unreleased version; fix by changing to LEIN_VERSION="2.9.4"

### DIFF
--- a/bin/lein
+++ b/bin/lein
@@ -8,7 +8,7 @@ function msg {
     echo "$@" 1>&2
 }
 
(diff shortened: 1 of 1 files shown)


winks:
This is intended, although maybe it could be messaged a bit better. 

https://github.com/technomancy/leiningen/blob/master/CONTRIBUTING.md#uberjar-from-master mentions `-SNAPSHOT` - if you want 2.9.4, checkout the 2.9.4 tag and it should work.

PR #2702, opened 2020-09-27
Make leiningen.core.project/meta-merge public

robhanlon22:
`meta-merge` is such a useful function that there's even a [standalone version of it available](https://github.com/weavejester/meta-merge). When writing Leiningen plugins that modify the project map, having this handy function available without having to spelunk private vars could be very useful.


technomancy:
Originally this was kept private to protect against changes in the implementation, but at this point that seems unlikely. Why not? =)

PR #2712, opened 2020-11-22
Update commons-io to 2.8

utkarsh2102:
Hi @technomancy,

With `commons-io` v2.8 in the Debian archive, the build
for `leiningen` in Debian is broken and hence this PR.

Signed-off-by: Utkarsh Gupta <<utkarsh@debian.org>>


technomancy:
Makes sense; thank you.

PR #2701, opened 2020-08-29
Changed dependency sjacket to trptcolin's fork

nighcoder:
Adds support for namespaced maps  in lein repl.
Fixes #2310.

PR #2713, opened 2020-12-07
Upgrade Apache httpclient

c2nes:
This upgraded version includes a patch for CVE-2020-13956. The version of httpcore has been updated to match the version preferred by httpclient itself (see [here](https://github.com/apache/httpcomponents-client/blob/rel/v4.5.13/pom.xml#L69)).

Regenerating `bootclasspath-deps.clj` according to the instructions in the file also picked up a group ID change for `sjacket` which appears to correspond to the change made in #2701.


technomancy:
Thanks for putting this together.

Will get a release put out soon.

Issue #2715, opened 2020-12-16
`lein test` loads namespaces out of order and multiple times


frenchy64:
**Initial debugging steps**
Before creating a report, _especially_ around exceptions being thrown when running Leiningen, please check if the error still occurs after:

- [x] Updating to using the latest released version of Leiningen (`lein upgrade`).
- [x] Moving your `~/.lein/profiles.clj` (if present) out of the way. This contains third-party dependencies and plugins that can cause problems inside Leiningen.
- [x] Updating any old versions of plugins in your `project.clj`, especially if the problem is with a plugin not working. Old versions of plugins like nREPL and CIDER (as well as others) can cause problems with newer versions of Leiningen.
- [x] (If you are using Java 9 or newer), updating your dependencies to their most recent versions. Recent JDK's have introduced changes which can break some Clojure libraries.

**Describe the bug**
`lein test` can reload namespaces out of order. This can leave deftypes implementing expired protocol interfaces, and thus an exception will be thrown when attempting to invoke a protocol method.

This is also a potential performance problem, since a namespace maybe reloaded twice.

**To Reproduce**
A possible cause of this bug is that `lein test` uses clojure.core/require's :reload flag.

Working backwards from there, a simple way to make `require :reload` reload things out of order is to call

```clojure
(require :reload 'A 'B)
```

where A depends on B. B will be loaded twice, first as a dependency of A, then via the `:reload` logic.

If B contains a protocol and A contains a deftype, then the deftype will implement a now-stale interface.

This scenario is demonstrated [here](https://github.com/frenchy64/lein-test-reload-bug), and can be triggered with `lein test`.

The concrete scenario involves 2 namespaces:

```clojure
(ns lein-test-reload-bug.a-deftype
  (:require [lein-test-reload-bug.b-protocol
             :refer [B]]))

(deftype A []
  B
  (b [this]))
```

```clojure
(ns lein-test-reload-bug.b-protocol)

(defprotocol B
  (b [this]))
```

Notice how this matches the hypothetical scenario, except
- A = lein-test-reload-bug.a-deftype
- B = lein-test-reload-bug.b-protocol

The following require is called by `lein test`:

```clojure
(require :reload 'lein-test-reload-bug.a-deftype
                 'lein-test-reload-bug.b-protocol)
```

**Actual behavior**
`lein-test-reload-bug.b-protocol` is loaded twice, the second time is _after_ `lein-test-reload-bug.a-deftype`, thus leaving it in a bad state.

[...]

PR #2716, opened 2020-12-16
Don't use :reload when loading namespaces in `lein test`

frenchy64:
Closes https://github.com/technomancy/leiningen/issues/2715

See linked issue for implications of using reload.

We have an exception for `:eval-in :nrepl` users, because they are (probably) the only ones who run `lein test` in a dirty environment where the `:reload` option makes a significant difference.

### DIFF
--- a/src/leiningen/test.clj
+++ b/src/leiningen/test.clj
@@ -73,17 +73,23 @@
                            ~selectors))]
       ns#)))
 
-;; TODO: make this an arg to form-for-testing-namespaces in 3.0.
+;; TODO: make this an option to form-for-testing-namespaces in 3.0.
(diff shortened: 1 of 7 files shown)


technomancy:
Thanks for starting on this.

For background, the reason for the `:reload` is that some people keep their repl server running and set Leiningen to run with `:eval-in :nrepl` so that a run of `lein test` will send the code to be evaluated in a running nrepl server rather than starting a brand new project process.

I think nowadays this mode of operation is less common, but it would be good if it still worked.

Maybe we can detect `:eval-in :nrepl` in the project map and retain the reloading behavior in that case?

technomancy:
Looks great; thanks for adding that in.

Issue #2718, opened 2020-12-17
lein test returns exit code 0 for multiples of 256 test failures+errors


frenchy64:
**Initial debugging steps**
Before creating a report, _especially_ around exceptions being thrown when running Leiningen, please check if the error still occurs after:

```
- [x] Updating to using the latest released version of Leiningen (`lein upgrade`).
- [x] Moving your `~/.lein/profiles.clj` (if present) out of the way. This contains third-party dependencies and plugins that can cause problems inside Leiningen.
- [x] Updating any old versions of plugins in your `project.clj`, especially if the problem is with a plugin not working. Old versions of plugins like nREPL and CIDER (as well as others) can cause problems with newer versions of Leiningen.
- [x] (If you are using Java 9 or newer), updating your dependencies to their most recent versions. Recent JDK's have introduced changes which can break some Clojure libraries.
```
**Describe the bug**
`lein test` returns a `0` exit code for multiples of 256 test failures+errors.

**To Reproduce**

```clojure
(defproject lein-test-exit-code "0.1.0-SNAPSHOT"
  :dependencies [[org.clojure/clojure "1.10.1"]])
```

```clojure
(ns lein-test-exit-code.core-test
  (:require [clojure.test :refer [deftest is]]))

(defmacro gen-failing-deftests [n]
  `(do 
     ~@(for [i (range n)]
         `(deftest ~(symbol (str "failing-" i))
            (is false)))))

(gen-failing-deftests 256)
```

**Actual behavior**
```
$ lein test
...
...
Ran 256 tests containing 256 assertions.
256 failures, 0 errors.

$ echo $?
0
```

**Expected behavior**
A non-zero exit code should be returned for test failures.

**Environment**
- Leiningen Version:
```
Leiningen 2.9.5 on Java 1.8.0_275 OpenJDK 64-Bit Server VM
```
- Leiningen installation method: manual
- JDK Version:
```
openjdk version "1.8.0_275"
OpenJDK Runtime Environment (AdoptOpenJDK)(build 1.8.0_275-b01)
OpenJDK 64-Bit Server VM (AdoptOpenJDK)(build 25.275-b01, mixed mode)
```
- OS: macOS 10.15

frenchy64:
Some extra context:
- exit codes wrap modulo 256 https://tldp.org/LDP/abs/html/exitcodes.html#AEN23629
[...]

PR #2719, opened 2020-12-17
WIP: Fix lein test returning 0 exit code on failure

frenchy64:
Closes https://github.com/technomancy/leiningen/issues/2718

This updates `lein test` to return exit code 0 on success and 1 on failure.

### DIFF
--- a/test/leiningen/test/test.clj
+++ b/test/leiningen/test/test.clj
@@ -3,6 +3,7 @@
   (:require [clojure.test :refer :all]
             [leiningen.test :refer :all]
             [leiningen.test.helper :refer [tmp-dir sample-no-aot-project
+                                           lein-test-exit-code-project
                                            lein-test-reload-bug-project
                                            sample-reader-cond-project
                                            sample-failing-project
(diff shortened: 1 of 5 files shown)


test/leiningen/test/test.clj
```diff
@@ -79,6 +80,15 @@
   ;; Issue #2715
   (test lein-test-reload-bug-project))
 
+(deftest test-failure-exit-code
```
frenchy64, line 83: Could I please get some help making a test that checks the correct exit code? I'm stuck.

See the CI logs for context, this test fails with:

```clojure
FAIL in (test-failure-exit-code) (test.clj:85)
expected: (= 1 (try (test lein-test-exit-code-project) false (catch clojure.lang.ExceptionInfo e (:exit-code (ex-data e)))))
  actual: (not (= 1 false))
```

technomancy:
Looks like there was some trouble introduced by the `*exit-after-tests* false` binding here; took care of that and it seems OK now.

Good catch; thanks!

PR #2741, opened 2021-05-05
Fix infinite loop when unpatched gpg used with openpgp keyservers

kipz:
Opengpg servers don't return user-ids when `gpg --recv-keys <key>` is run, but `gpg` returns 0 exit code but does _not_ import the key, so Leiningen assumes success and enters an infinite loop.

Some background:
https://keys.openpgp.org/about/faq#older-gnupg
https://dev.gnupg.org/T4393#133689


technomancy:
Thanks!

I'm not sure why CI skipped this PR but running it locally I get a test
failure because it says commons-io is :signed instead of :no-key.

I've removed that check and applied the patch; thanks again.

technomancy:
(GitHub didn't detect that I merged your patch so I had to close it manually but it has actually landed on the main branch despite looking otherwise)

PR #2742, opened 2021-05-05
Setup GNUPGHOME for :verify test. Add commons-io/:no-key to assertion

kipz:
As discussed, here's an attempt at setting up the environment for the test we added in #2741 

I thought it safest to add a `redef` for this particular test rather than try change the environment for all tests - I'm a little worried that it might break other local workflows with different gpg versions/configs like I did before.


technomancy:
Looks great; thanks.

PR #2755, opened 2021-08-19
Update REPLy to 0.5.0

bbatsov:
This version of REPLy replaces clojure-complete with the newer
incomplete library. More details - https://github.com/nrepl/incomplete

The end users shouldn't be affected negatively by the change, they'll just get better completion results.

### DIFF
--- a/doc/PROFILES.md
+++ b/doc/PROFILES.md
@@ -260,8 +260,8 @@ plugin:
      :group "lein-pprint",
      :source-path ("/home/phil/src/leiningen/lein-pprint/src"),
      :dependencies
-     ([org.clojure/tools.nrepl "0.0.5" :exclusions [org.clojure/clojure]]
-      [clojure-complete "0.1.4" :exclusions [org.clojure/clojure]]
+     ([nrepl "0.8.3" :exclusions [org.clojure/clojure]]
+      [incomplete "0.1.0" :exclusions [org.clojure/clojure]]
       [org.thnetos/cd-client "0.3.3" :exclusions [org.clojure/clojure]]),
      :target-path "/home/phil/src/leiningen/lein-pprint/target",
      :name "lein-pprint",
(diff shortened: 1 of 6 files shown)


technomancy:
Looks good; thanks.

technomancy:
This causes errors for me on a fresh project:

```
$ lein repl
nREPL server started on port 34843 on host 127.0.0.1 - nrepl://127.0.0.1:34843
REPL-y 0.5.0, nREPL 0.8.3
Clojure 1.10.3
OpenJDK 64-Bit Server VM 11.0.11+9-post-Debian-1deb10u1
    Docs: (doc function-name-here)
          (find-doc "part-of-name-here")
  Source: (source function-name-here)
 Javadoc: (javadoc java-object-or-class-here)
    Exit: Control+D or (exit) or (quit)
 Results: Stored in vars *1, *2, *3, an exception in *e

hey.core=> clojure.string/reExecution error (ArityException) at incomplete.core/scoped-candidates (core.clj:280).
Wrong number of args (2) passed to: reply.exports/eval2059/resolve-class--2060
```

bbatsov:
@technomancy I'll take a look. I hadn't notice that REPLy was using somewhere `resolve-class` directly, but this should be an easy fix.

bbatsov:
The problem is here https://github.com/trptcolin/reply/blob/14b11afb07153b51005dcc4a23de45928f9ed7ff/src/clj/reply/initialization.clj#L153

In REPLy there's some definition of `resolve-class` taking only a symbol and in incompletely it's a symbol and its namespace. I'll probably just make a small change to in incomplete to work with only a symbol as well today, and I'll consult with @trptcolin for the best change in REPLy, as I can't find where this export is actually being used.

bbatsov:
I've submitted a proposed fix to REPLy here https://github.com/trptcolin/reply/pulls/206

technomancy:
Thanks for following up. I'll keep an eye out for that fix to land and pull it in when it does.

bbatsov:
The ETA for a new release is later today. I've just addressed a bit of feedback @trptcolin had about my PR.  

On a positive note - I've fixed a couple of unrelated issues in REPLy as well, so at least we'll get a bit more out of 0.5.1. :-)

trptcolin:
0.5.1 is out, sorry for the issue!

bbatsov:
I've opened a PR updating REPLy to 0.5.1 https://github.com/technomancy/leiningen/pull/2762

Issue #2251, opened 2017-03-02
Don't warn about version ranges when the range is "[a,a]"


danielcompton:
In #734, lein added warning on dependency version ranges. This makes sense in the general case. However I noticed when running https://github.com/capitalone/cqrs-manager-for-distributed-reactive-services, that I got the following warning:

```
WARNING!!! version ranges found for:
[io.grpc/grpc-netty "1.0.0"] -> [io.netty/netty-codec-http2 "[4.1.3.Final,4.1.3.Final]"]
Consider using [io.grpc/grpc-netty "1.0.0" :exclusions [io.netty/netty-codec-http2]].
[io.grpc/grpc-netty "1.0.0"] -> [io.grpc/grpc-core "[1.0.0,1.0.0]"]
Consider using [io.grpc/grpc-netty "1.0.0" :exclusions [io.grpc/grpc-core]].
```

If you look at the [pom](http://central.maven.org/maven2/io/grpc/grpc-netty/1.0.0/grpc-netty-1.0.0.pom) for grpc-netty, you can see:

```xml
<dependency>
  <groupId>io.netty</groupId>
  <artifactId>netty-codec-http2</artifactId>
  <version>[4.1.3.Final]</version>
  <scope>compile</scope>
</dependency>
```

I'm guessing, but suspect that this is is a pseudo version range, to stop Maven from selecting anything else. This is to aid in reproducibility, not hinder it. Leiningen sees this as `[4.1.3.Final,4.1.3.Final]` and warns on the version range.

Does it make sense to disable the warning in this specific case where the range specifier is the same on both ends?

hyPiRion:
If pedantic only were there to ensure repeatability, I'd agree. However, it's also there to help out in case you don't understand why your version isn't picked up.

In this case, I'd like to get this warning if I try to override the version of `io.netty/netty-codec-http2`, and would be very confused if my `[io.netty/netty-codec-http2 "4.1.8.Final"]` vector did absolutely nothing (Or even worse, I wouldn't detect it and deploy it with `"4.1.3.Final"`).

If it's possible to detect whether a pinned (`[x]`) version would overrun some other dependency on the same artifact, then I'm fine with relaxing the warnings when it doesn't affect the outcome.

emlyn:
I can see why the warning cannot just be removed in this case, but I still think something needs to be done.

[...]

PR #2757, opened 2021-08-24
Exclude ranges which point to a single version from pedantic

pyr:
Pedantic rightly complains when version ranges introduce unclear
behavior. Some projects always use ranges and have the same
lower and upper bound when pointing to a single version.

This patch captures this behavior to avoid suggesting exclusions
in this case.

Any project using GRPC artifacts experiences this for instance.
I must admit I don't have a clear idea on how to implement tests for
this patch though

### DIFF
--- a/leiningen-core/src/leiningen/core/pedantic.clj
+++ b/leiningen-core/src/leiningen/core/pedantic.clj
@@ -42,10 +42,16 @@
         (.transformGraph node context))))
 
 (defn- range?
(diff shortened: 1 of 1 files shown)


pyr:
After reading #2251 this PR will need an additional bit of work to warn when a specific version is shadowed by a range like this, I will rework the patch to adapt it.

technomancy:
Seems pretty sensible! I'll accept the manual test as being sufficient. Thank you.

Issue #2764, opened 2021-09-16
Build on CircleCI fails with "Pathological dependency tree detected; consider disabling pedantic."


gixxi:
*Leiningen Version* 

2.9.7

works with *2.9.6*

*Description*

coincidence with latest change "Warn when pedantic detects a pathological dependency tree."

Is there a way to disable this feature 

[project.clj.txt](https://github.com/technomancy/leiningen/files/7174703/project.clj.txt)

gixxi:
happens with the project.clj

`(defproject rocklog-vlic "21.01-SNAPSHOT-20210916-0444"
  :description "Rocklog WMS based on the Very Large Intralogistics Kernel (VLIC) and Lambdaroyal-Memory"
  :url "https://github.com/lambdaroyal/rocklog-vlic"
  :license {:name "Proprietary"
            :url "http://www.planet-rocklog.com"}
  :dependencies [[org.clojure/clojure "1.10.3"]
                 [org.clojure/spec.alpha "0.2.176"]
                 [org.clojure/tools.cli "0.3.3"]
                 [org.clojure/core.async "0.4.490"]
                 [org.clojure/data.priority-map "0.0.7"]
                 [com.ashafa/clutch "0.4.0"
                  :exclusions [clj-http]]
                 [org.clojure/tools.logging "0.3.1"]
                 [org.slf4j/slf4j-log4j12 "1.7.1"]
                 [org.lambdaroyal/clojure-algorithms "0.2-SNAPSHOT"]
                 [org.lambdaroyal/clojure-util "1.0-SNAPSHOT"]
                 [org.clojars.gixxi/lambdaroyal-memory "1.1.5Beta"]
                 [javax.servlet/servlet-api "2.5"]
                 [ring/ring-json "0.5.0-beta1"]
                 ;;lib for caching (IP location etc.)
                 [org.clojure/core.memoize "0.5.6"]
                 ;;getting geolocations
                 [com.oscaro/maximator "0.1.0"]
                 [com.fasterxml.jackson.core/jackson-core "2.4.2"]
                 [com.fasterxml.jackson.core/jackson-databind "2.4.2"]

                 ;;read-in worksheets
                 [org.apache.commons/commons-vfs2 "2.0"]
                 [commons-lang "2.6"]

                 ;;accepting nRepl Connections
                 [org.clojure/tools.nrepl "0.2.12"]

                 ;;cerberrus
                 [schejulure "1.0.1"]
[...]

PR #2768, opened 2021-09-21
Exit w/ non-zero status on lein help not-found

### LINKED ISSUE #2767: lein help non-existant-task should exit with non-zero error code
You all might have good reasons for not doing this, but just in case I thought I'd ask:

Would it make sense to have lein exit with a non-zero error code when you run `lein help non-existant-subcommand-or-task`?

I ask because I'm working on enabling lein to be the ENTRYPOINT of the official clojure Docker images (instead of the CMD) but I also have to detect when the user is trying to invoke a different executable in the image directly and override that (due to Docker Official Image policy). The usual approach they take is to see if asking for help on that subcommand exits successfully or not (to detect if it's something that should be exec'd directly instead of passed to the ENTRYPOINT). But lein help currently appears to exit successfully (with `Task: 'whatever' not found` in this case) no matter what.

cap10morgan:
Closes #2767

### DIFF
--- a/src/leiningen/help.clj
+++ b/src/leiningen/help.clj
@@ -112,7 +112,7 @@
                 (subtask-help-for task-ns task)
                 (if (some seq (get-arglists task))
                   (str "\n\nArguments: " (pr-str (get-arglists task))))))
-         (format "Task: '%s' not found" task-name))))
+         (main/abort (format "Task: '%s' not found" task-name)))))
   ([project task-name]
      (let [aliases (merge main/aliases (:aliases project))]
        (or (alias-help aliases task-name)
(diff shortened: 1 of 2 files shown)


technomancy:
Looks great; thank you!

cap10morgan:
@technomancy Thanks for merging. I know you just had a release, but any chance on another one soon with this included? I have an open PR for the change to the official clojure Docker images I mentioned in #2767 that would be great to take advantage of this behavior in. I understand if this is a bit too minimal of a change to warrant another release so soon, though.

technomancy:
> I know you just had a release, but any chance on another one soon with this included?

I would like to cut another release but in order to do that I need more data around these two issues:

* https://github.com/technomancy/leiningen/issues/2769
* https://github.com/technomancy/leiningen/issues/2764

I have asked the reporters for more details in order to understand the problem better but I am still waiting.

Issue #2777, opened 2021-12-22
Opaque `java.lang.Exception: Couldn't create directories: ` error


vemv:
Forwarding from https://github.com/clojure-emacs/cider/issues/3108

The following stacktrace has been observed:

```
java.lang.Exception: Couldn't create directories: 
 at leiningen.core.utils$mkdirs.invokeStatic (utils.clj:72)
    leiningen.core.utils$mkdirs.invoke (utils.clj:67)
    leiningen.core.eval$prep.invokeStatic (eval.clj:83)
    leiningen.core.eval$prep.invoke (eval.clj:73)
    leiningen.core.eval$eval_in_project.invokeStatic (eval.clj:364)
    leiningen.core.eval$eval_in_project.invoke (eval.clj:358)
    clojure.lang.AFn.applyToHelper (AFn.java:160)
    clojure.lang.AFn.applyTo (AFn.java:144)
    clojure.core$apply.invokeStatic (core.clj:667)
    clojure.core$apply.invoke (core.clj:660)
    leiningen.repl$repl.invokeStatic (repl.clj:393)
    leiningen.repl$repl.doInvoke (repl.clj:319)
    clojure.lang.RestFn.applyTo (RestFn.java:142)
    clojure.lang.Var.applyTo (Var.java:705)
    clojure.core$apply.invokeStatic (core.clj:667)
    clojure.core$apply.invoke (core.clj:660)
    leiningen.core.main$partial_task$fn__7356.doInvoke (main.clj:284)
    clojure.lang.RestFn.applyTo (RestFn.java:139)
    clojure.lang.AFunction$1.doInvoke (AFunction.java:31)
    clojure.lang.RestFn.applyTo (RestFn.java:137)
    clojure.core$apply.invokeStatic (core.clj:667)
    clojure.core$apply.invoke (core.clj:660)
    leiningen.core.main$apply_task.invokeStatic (main.clj:334)
    leiningen.core.main$apply_task.invoke (main.clj:320)
    leiningen.core.main$resolve_and_apply.invokeStatic (main.clj:343)
    leiningen.core.main$resolve_and_apply.invoke (main.clj:336)
    leiningen.update_in$update_in.invokeStatic (update_in.clj:37)
    leiningen.update_in$update_in.doInvoke (update_in.clj:24)
    clojure.lang.RestFn.applyTo (RestFn.java:146)
    clojure.lang.Var.applyTo (Var.java:705)
[...]

PR #2778, opened 2021-12-22
Make `utils/mkdirs` nil-safe

vemv:
> Part of https://github.com/technomancy/leiningen/issues/2777

If for whatever reason `(utils/mkdirs nil)` was invoked, one would get `java.lang.Exception: Couldn't create directories:` which is not a very informative error.

I'd propose to make this defn's API akin to `io/file`'s - which returns nil on nil inputs instead of throwing.

Unfortunately I'm not able to reproduce the bug that is causing these issues in the first place, namely https://github.com/clojure-emacs/cider/issues/3108. So I can't debug why `(utils/mkdirs nil)` was being invoked to begin with.


technomancy:
Oops; haha, I guess I should read all the messages before replying.

Looks good; thanks.

PR #2779, opened 2021-12-22
Extend "lein change" to be able to edit versions in :dependencies and :managed-dependencies

ejschoen:
Consider an application suite with parent and child projects (i.e., using lein-parent) and with a versioning policy that requires, for example, consistently bumping the patch version of the parent project and the versions a collection of libraries or applications when a change is made to a common shared library.  

It was already possible to use `lein change` to bump the version of the parent project, and to bump the version of the reference to the parent project in the children (e.g., `lein change :parent-project:coords leiningen.release/bump-version :patch` followed by `lein change :parent-project:coords leiningen.release/bump-version :release`)

However, it was not possible to automate updating version references in the :dependencies or :managed-dependencies vectors of projects or parent projects.  So, for example upon the release of a hotfix to a shared jar that's built into multiple uberjars, it was necessary to manually edit the :managed-dependency vector of the parent project to update the version of the shared jar.  

This PR adds to leiningen.change/change the ability to parse the sjacket DOM for vectors of vectors, and to match on the first element of an inner vector, as in:
```
:dependencies [[org.clojure/clojure "1.8.0"]
               [org.clojure/tools.cli "0.4.2"]]
```
and by generalizing slightly leiningen.change/normalize-path to handle specifications such as:

```
lein change :dependencies[org.clojure/clojure] set '"1.10.1"'
```

The above works in :dependencies and in :managed-dependencies vectors in project.clj files, and enables developers and devops processes to script the updates to version dependencies in project files.

### DIFF
--- a/src/leiningen/change.clj
+++ b/src/leiningen/change.clj
@@ -22,12 +22,28 @@
 ;; NOTE: this destroy comments, formatting, etc.
 (defn- sjacket->clj [value]
   (if-not (#{:comment :whitespace :newline} (:tag value))
-    (-> value sj/str-pt read-string)))
+    (try (-> value sj/str-pt read-string)
+         (catch Exception e 
+           (println "****" (.getMessage e))
+           (println value)
+           (throw e))) ))
 
 (defn ^:internal normalize-path [value]
   (if (coll? value)
     value
-    (map keyword (remove empty? (str/split value #":")))))
+    (->> (re-seq #"(?:[:\[])?(?:[^:\[]*)?" value)
+         (remove empty?)
+         (map (fn [element]
+                (cond (re-matches #"\[.+" element)
+                      (if (re-matches #".+\]" element)
+                        (read-string (subs element 1 (dec (count element))))
+                        (throw (Exception. (format "Bad path element %s.  Must encode group/artifact name in []" element)))) 
+                      (re-matches #":.*" element)
+                      (if (not= element ":")
+                        (keyword (subs element 1))
+                        nil) 
+                      :else (keyword element))))
+         (filter identity))))
 
 (defn ^:internal collapse-fn [f args]
   (let [f (cond (ifn? f) f
@@ -80,10 +96,23 @@
 
 (defn- find-right [loc pred]
   (->> loc
-      (iterate zip/right)
-      (take-while (comp not nil?))
-      (filter (comp pred zip/node))
-      first))
+       (iterate zip/right)
+       (take-while (comp not nil?))
+       (filter (comp pred zip/node))
+       first))
+
+(defn- find-dependency-version [loc groupid-artifactid]
+  "Find the entry in a vector of dependencies or managed dependencies whose first element
+   matches the symbol bound to groupid-artifactid.  Return the loc of the first string
+   in that entry, which we assume to be a version number."
+  (-> (->> loc
+           (iterate zip/right)
+           (take-while (comp not nil?))
+           (remove insignificant?)
+           (filter (comp #{groupid-artifactid} first sjacket->clj first))
+           first)
+      zip/down
+      (find-right (comp #{:string} :tag))))
 
 (defn- find-key [loc key]
   (->> loc
@@ -125,6 +154,7 @@
       (zip/edit (comp clj->sjacket fn sjacket->clj))
       zip/root))
 
+
 (defn- update-name [proj fn]
   (-> proj
       zip/right
@@ -134,7 +164,9 @@
       zip/root))
 
 (defn- update-setting [proj [p & ath] fn]
-  (let [loc (or (-> proj (find-key p) next-value)
+  (let [loc (or (cond (keyword? p) (-> proj (find-key p) next-value)
+                      (symbol? p) (-> proj (find-dependency-version p))
+                      :else (throw (Exception. (format "Invalid update-setting path component: %s" (pr-str p)))))
                 (-> proj
                     zip/rightmost
                     (insert-entry p)
@@ -170,8 +202,9 @@ well as turning string args into Clojure data; this function handles the rest."
 (defn change
   "Rewrite project.clj with f applied to the value at key-or-path.
 
-The first argument should be a keyword (or mashed-together keywords for
-nested values indicating which value to change). The second argument
+The first argument should be a keyword,  mashed-together keywords for
+nested values, or keyword(s) separated by : and ending in a symbol enclosed in square braces,
+indicating which value to change). The second argument
 should name a function var which will be called with the current value
 as its first argument and the remaining task arguments as the rest.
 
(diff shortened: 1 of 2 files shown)


technomancy:
Thanks for submitting this patch.

I think this functionality is a good idea, but I think we should offer a better notation for it. The problem with the proposed notation is that previously, all the arguments are just in normal Clojure data syntax, and this patch sacrifices the consistency we once had.

A better approach would be to provide the behavior you want as a function. There is no need to change the argument-parsing code at all.

Rather than this: `lein change :dependencies[org.clojure/clojure] set '"1.10.1"'`

We should use this: `lein change :dependencies leiningen.change/set-version [org.clojure/clojure] '"1.10.1"'`

Does that make sense?

ejschoen:
I had implemented exactly that first.  But the problem is that leiningen.change/set-version would be passed a Clojure vector rather than an sjacket DOM value.  So while it was possible to edit the version of a specific dependency, the resulting project.clj lost its original formatting, and that's not OK.  I didn't see any alternative but to work at the level of the sjacket DOM traversal.

I'd be open to implementing something that parses the smashed path syntax without breaking existing scenarios, which is what the current PR does.  But yes, it's a different syntax.  I was trying to come up with a syntax that seemed natural to a programmer, but I agree it's more Java-like than Clojure-like.  

I also thought about making the syntax for the path be like `:dependencies:org.clojure/clojure', but that's not really Clojure data syntax either, and the `normalize-path` function would have figure out to instantiate a keyword or a namespaced symbol.  

Then I thought about implementing a different entry point: `lein change-dependency ...` but I'd end up having to reimplement a lot of `lein change` if I did that.

Is it true to characterize the current implementation as taking all arguments as Clojure data syntax?  For example, we use this frequently:

`lein change :parent-project:coords set '[group/artifact "version"]'`

where `:parent-project:coords` isn't really normal Clojure data syntax.

technomancy:
Ah, I see what you mean, yeah the distinction between raw Clojure data vs the formatting-aware straightjacket data complicates things.

> Is it true to characterize the current implementation as taking all arguments as Clojure data syntax?

No, you're right--it's been a while since I'd worked in this code and I forgot that we do allow nested keywords to be concatenated together without any spacing due to limitations of how the shell passes arguments thru. So that is one concession made.

> I also thought about making the syntax for the path be like
> `:dependencies:org.clojure/clojure', but that's not really Clojure
> data syntax either, and the `normalize-path` function would have
> figure out to instantiate a keyword or a namespaced symbol.
 
I actually like this a lot more! It retains the existing concession of leaving out the space, but if you just imagine the space is there, it lets you pretend that the `:dependencies` list is a map rather than a vector.

In fact; this is legal in `project.clj`; the dependencies are typically given as a seq of vectors because ordering is important, but if you don't care about ordering, Leiningen will still accept a map since calling `seq` on that map will get you a list of vectors anyway.

```
  :dependencies {org.clojure/clojure "1.10.1"                                                                                                       
                 ring "1.8.2"}
```

The only downside is that you have to encode in the `change` task that certain keys are treated specially. But this goes "with the grain" of the data given the relationship between maps and vectors, more than the proposed syntax IMO.

ejschoen:
OK.  The smashed keyword syntax now works:

```lein change :dependencies:org.clojure/clojure set '"1.10.1"'```

The `update-setting` function now determines whether the sjacket DOM location corresponding to the current path element is a map or vector, and interprets the next path element accordingly.  

For completenes, the code can now append a dependency for a group/artifact that isn't in the dependency or managed-dependency vector when used with the set function.

File: src/leiningen/change.clj [hunk 1]
```diff
@@ -133,15 +153,38 @@
...
+                  (iterate zip/right)
+                  (take-while (comp not nil?))
+                  (remove insignificant?)
+                  (filter (comp #{:string :map :vector} :tag zip/node))
+                  first
+                  zip/node)]
+    (:tag node)))
+
+(defn- update-setting [proj datatype [p & ath] fn]
+  (let [loc (or (case datatype
```
technomancy, line 167: I feel like this function is getting a bit out of hand. Could you split it out and make it so each `case` is its own function so that the names of the functions can help give you a better hint of what's going on?

src/leiningen/change.clj
```diff
@@ -80,10 +83,26 @@
...
 (defn- find-right [loc pred]
   (->> loc
-      (iterate zip/right)
-      (take-while (comp not nil?))
-      (filter (comp pred zip/node))
-      first))
+       (iterate zip/right)
+       (take-while (comp not nil?))
+       (filter (comp pred zip/node))
+       first))
```
technomancy, line 89: Looks like the indentation is off here.

src/leiningen/change.clj
```diff
@@ -21,11 +21,14 @@
 
 ;; NOTE: this destroy comments, formatting, etc.
 (defn- sjacket->clj [value]
-  (if-not (#{:comment :whitespace :newline} (:tag value))
-    (-> value sj/str-pt read-string)))
+  (if-not (or (#{:comment :whitespace :newline} (:tag value))
+              (#{"]" "}"} value))
+    (try (-> value sj/str-pt read-string)
+         (catch Exception e 
+           (throw e))) ))
```
technomancy, line 28: This `try/catch` doesn't seem to have any effect and should probably be removed, right?

src/leiningen/change.clj
```diff
@@ -165,13 +208,14 @@ well as turning string args into Clojure data; this function handles the rest."
...
-       (update-setting (zip/right proj) path f)))))
+       (update-setting (zip/right proj) :map path f)))))
 
 (defn change
   "Rewrite project.clj with f applied to the value at key-or-path.
 
-The first argument should be a keyword (or mashed-together keywords for
-nested values indicating which value to change). The second argument
+The first argument should be a keyword,  mashed-together keywords for
+nested values, or keyword(s) separated by : and ending in a symbol enclosed in square braces,
```
technomancy, line 217: Let's try to keep the docstrings under 80 columns; thanks.

test/leiningen/test/change.clj
```diff
@@ -118,14 +118,29 @@
...
+  (testing "can set dependency version"
+    (is (= "(defproject leiningen.change \"0.0.1\" :description \"a dynamic description\" :dependencies [[org.clojure/clojure \"1.10.1\"]])"
+           (change-string "(defproject leiningen.change \"0.0.1\" :description \"a dynamic description\" :dependencies [[org.clojure/clojure \"1.8.0\"]])"
+                          ":dependencies:org.clojure/clojure"  "set" "1.10.1")
+           )))
+  (testing "can append dependency version"
+    (is (= "(defproject leiningen.change \"0.0.1\" :description \"a dynamic description\" :dependencies [[org.clojure/clojure \"1.8.0\"] [org.clojure/core.cache \"1.0.207\"]])"
+           (change-string "(defproject leiningen.change \"0.0.1\" :description \"a dynamic description\" :dependencies [[org.clojure/clojure \"1.8.0\"]])"
+                          ":dependencies:org.clojure/core.cache"  "set" "1.0.207")
+           ))))
```
technomancy, line 131: Can we add a test for a case where the target dependency has more than just a version? Like a dependency vector that has a `:classifier` or similar setting in it.

Also if we could avoid parens on their own lines that'd be great.

technomancy (commented):
Thanks; this looks a lot better. Got a few notes for some cleanup but overall this feels like it's on the right track.

ejschoen:
Formatting cleaned up.  Test added for dependencies with e.g., exclusions, and update-setting refactored to use a polymorphic helper function that dispatches based upon whether the path descends into a map or vector.

technomancy:
Thanks; that's much better!

PR #2791, opened 2022-04-12
Add aarch64 support to leiningen.core.utils

otremblay:


technomancy:
Thanks; applied in 60f7ab7.

PR #2792, opened 2022-04-21
print newest deps

cdzwm:
When conflict deps found, print the depency of newest version.


technomancy:
Thanks for this patch.

I think overall it's good to surface this information, but the way it's presented in this implementation is not particularly clear. Simply saying `Suggest: ...` is confusing because it's presented alongside the suggestion to use `:exclusions` and it's not obvious how the two different suggestions are related. I think this needs some more thought put into it, but we could include it if the information is presented in a less confusing way.

There's also a couple technical issues with it; for one it removes `:eval-in :leiningen` which is pretty important, so that needs to be changed.

Also it appears to bring in a new dependency, which causes a lot of extra work for downstream packagers. As far as I can tell, this is only being used for a single function, which duplicates logic already available from Aether: https://github.com/technomancy/leiningen/blob/master/leiningen-core/src/leiningen/core/pedantic.clj#L68 Let's use the existing implementation instead of bringing in something redundant.

cdzwm:
According to you suggestion, I make some modifications.

technomancy:
Sorry, I haven't gotten a chance to look at this yet, but I don't think it needs to be closed. Was that intentional?

cdzwm:
II'd like to think about this modification again. I think there are still some problems, but you can also take a look at it first and suggest some changes.

cdzwm:
Fix is completed.

technomancy:
Sorry about this; for some reason github did not notify me about the changes here. This is a lot better. There's still a few issues I'd like to see fixed.

```
Possibly confusing dependencies found:
[clj-kondo "2022.12.10"] -> [org.babashka/sci "0.5.35"] -> [borkdude/edamame "1.0.0"]
 overrides
[metosin/reitit "0.7.2"] -> [metosin/reitit-malli "0.7.2"] -> [metosin/malli "0.16.4"] -> [borkdude/edamame "1.4.27"]

Consider using these exclusions:
[metosin/reitit "0.7.2" :exclusions [borkdude/edamame]]

OR

Adding the next line to the head of the project.clj file's depency vector:
[borkdude/edamame "1.4.27"]


In addition to using above exclusion method, you can also add all the following lines
to the head of the depency vector of your project.clj to resolve the confusing dependencies' problem:

[borkdude/edamame "1.4.27"]
----------------------------------------
```

* Running the tests shows a bunch of reflection warnings now
* It seems to repeat the same suggestion twice with different wording
* Misspelling of dependency
* No tests

PR #2793, opened 2022-04-28
Remove README link to nonexistent plugins page

pfernandez:
There may once have been a list of plugins, but the link now points to a blank page.


technomancy:
It looks like someone (maliciously or by accident) deleted the plugins page on the wiki.

Instead of merging this PR, I've reverted the deleted page. Thanks for bringing it to my attention.

pfernandez:
Even better, thanks!

PR #2799, opened 2022-06-18
Add "repl :headless :socket PATH" support

rlbdv:
Just for review, I've structured the commits so that the socket test can be run by installing current nrepl 1.0.0-SNAPSHOT, checking out the "Update nrepl to 1.0.0-SNAPSHOT for test-headless-socket" commit, and running `lein test :only leiningen.test.repl/test-headless-socket`.  (Some of the existing tests don't appear to be compatible with the new nrepl.)

The subsequent commits just revert the nrepl version change, and mark that test as `:disabled`.  Before merging, I suspect we may want to squash everything into the first commit, and I'd be happy to do so.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,7 @@
 * Fix an issue with `check` where AOT would shadow reflection warnings. (Phil Hagelberg)
 * Allow `change` to edit dependency versions. (Eric Schoen)
 * Fix a bug where composite profiles would leak dependencies downstream. (Phil Hagelberg)
+* Allow `repl` to bind to filesystem sockets via `:headless :socket PATH` (Rob Browning)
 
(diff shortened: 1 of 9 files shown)


technomancy:
Thanks; I've merged this in b676f29. We can re-enable the test once the new nrepl is released.

bbatsov:
Might be good to also mention this somewhere in the docs.

technomancy:
@bbatsov it's mentioned in the help text for the repl task and also in `sample.project.clj`; is there somewhere else that you had in mind?

bbatsov:
Ah, I totally missed the `sample.project.clj` change. I guess we're good - I just wanted to make sure the end users will be able to find this.

Issue #2811, opened 2024-01-15
Global :exclusions in profile do not apply to root :dependencies


jpmonettas:
Currently adding a global `:exclusions` in a profile doesn't remove dependencies added directly in the main `:dependencies` vector.

This is useful when you want your personal profile to remove a dependency for development without having to comment out dependencies in your root vector.

For a concrete example, for using  ClojureStorm with [FlowStorm](http://www.flow-storm.org/) you need to be able to swap the official Clojure compiler by the ClojureStorm dev compiler.

The way to do this is to configure your project.clj like this : 
```clojure
(defproject my.project "1.0.0"
  :dependencies [
                ;; [org.clojure/clojure "1.11.1"]
                ]
  :profiles {:dev {:dependencies [[com.github.flow-storm/clojure "1.11.1-17"]]
                   :exclusions [org.clojure/clojure]}})
```   
where you exclude `org.clojure/clojure` to include `com.github.flow-storm/clojure` which provides a compiler replacement under a different group. The problem is this doesn't work unless you comment out the `org.clojure/clojure` dependency.

**To Reproduce**
1. Create an empty folder with the project.clj file described above
2. Run lein repl. The lein welcome message should say `Clojure 1.11.1-17`
3. Uncomment the `org.clojure/clojure` dependency while keeping  the exclusion
4. Run lein repl. The lein welcome message now says `Clojure 1.11.1`, which means the exlcusion did not apply

This is independent of lein, OS, and jvm version

technomancy:
OK, sounds reasonable. I would take a patch for this.

jpmonettas:
@technomancy The problem is related to how `[org.clojure/clojure "1.11.1" :exclusions [[org.clojure/clojure]]]` is handled.

For building the `java -classpath ...` command for starting the JVM process lein is currently calling`leiningen.core.classpath/get-classpath` with a project that contains :dependencies `[... [org.clojure/clojure "1.11.1" :exclusions [[org.clojure/clojure]]]]` which in turn calls `cemerick.pomegranate.aether/dependency-files` which doesn't take into account exclusions (it just returns the jar file on the coordinates meta).

So my question is, should something like `[org.clojure/clojure "1.11.1" :exclusions [[org.clojure/clojure]]]` be there in the first place? Should it be discarded as early as possible (like by `leiningen.core.project/read`), or we want to keep that and discard it before building a classpath?

technomancy:
[...]

PR #2813, opened 2024-01-22
Fix global :exclusions in profile not applying to root :dependencies

jpmonettas:
This fixes https://github.com/technomancy/leiningen/issues/2811

I ended adding a `remove-self-excluded-deps` instead of modifying `add-global-exclusions` to `add-and-apply-global-exclusions` because I think  `add-and-apply-global-exclusions` conveys that it will somehow apply the exclusions to the deps tree there which is misleading.

### DIFF
--- a/leiningen-core/src/leiningen/core/project.clj
+++ b/leiningen-core/src/leiningen/core/project.clj
@@ -465,6 +465,26 @@
                         (meta dependencies)))
       project)))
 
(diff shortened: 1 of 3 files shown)


jpmonettas:
The failures and errors shown in CI seems to be the same showing on the previous commit so I don't think they are related to this PR.

technomancy:
Nice work; this should do the trick. Thank you!

PR #2829, opened 2025-08-15
Upgrade clojure to 1.11.4

### LINKED ISSUE #2828: Bump clojure version
This is a restatement of #2816.  Our institutional rules make CVE's difficult to get exceptions, so we're still having problems because of this vulnerability (regardless of whether it's a nothing-burger).  It's also the case that lein is several releases back on clojure (1.11.1, if I'm reading it right; the 1.11.x line is at 1.11.4, and clojure 1.12.1 is already out), so I'd hope that an upgrade might be justified on its own terms.  It'd be a massive help to our project to have lein bump up versions.

jpolchlo:
Simple PR to update the version of clojure to address annoying security findings.

Closes #2828


jpolchlo:
I added the POM because it seemed like the right thing to do.  I noted that during testing a GPG file was updated, but I'm not going to include it.  That file seems like it should be removed from the repo and added to the .gitignore.  I can add that to this PR if that's a good thing to do.

technomancy:
Thanks. I've applied your commits but github didn't realize they were part of this PR for some reason, so I'm closing manually, but the changes were accepted.

PR #2831, opened 2026-02-26
Add "--bail" flag to lein test to stop execution early on failures

marco-carvalho:
## Problem
When a test suite has many namespaces, developers waste time waiting for the entire run to finish even after the first failure is already detected.
## Solution
Introduce a "--bail" flag that stops running remaining test namespaces once the accumulated failure count reaches a configurable threshold, giving developers faster feedback.
## Examples
```sh
# Stop after the first failure
lein test --bail

# Stop after 3 failures
lein test --bail 3

# Combine with a specific namespace
lein test my.app.core-test --bail

# Combine with selectors
lein test --bail :integration
```


technomancy:
Thanks for submitting this feature.

In general, my position towards the `test` task has been to avoid adding features, because it's better to handle this in a 3rd-party library, as indicated in the docstring:

> If you are looking for more control over your tests, consider 3rd-party runners like Kaocha.

In this particular case, the feature is already supported by Kaocha: https://cljdoc.org/d/lambdaisland/kaocha/1.91.1392/doc/cli-fail-fast-option

However, I'm not necessarily against adding this, but I'd like to know more about the context. Do you have a reason to prefer the built-in test task over something like Kaocha? In my opinion as its author, the built-in task is not very good.

marco-carvalho:
Seems fair, gonna close it. Thanks!

